Issue #64, opened 2015-03-18
Percentage scores now invisible


mjshepherd:

mjshepherd:
phenotips/phenotips/issues/1502

veronikaslc:
Closed to the favor of https://github.com/phenotips/patient-network/pull/102

PR #65, opened 2015-03-18
Issue #64: Percentage scores now invisible

mjshepherd:


buske:
Bump.

sdumitriu:
I don't understand this. It seems to be copy/pasted code from `SimilarCases`, and other than a slightly smaller font size and a hardcoded black border, there are no changes. Why?

buske:
@sdumitriu I believe this more tightly constrains the scope of these css definitions, so they don't accidentally affecting those that govern the similarcases variant display.

mjshepherd:
@sdumitriu @buske The code is exactly the same except for a slight difference in size and colour. I wasn't sure what the best practice would be in this case. I didn't have to copy every style, I could have just changed the size and colour in the more specific selector. I duplicated the code because I felt that changes to either the similar cases styles or the top exomiser styles should not effect the appearance of the other.

buske:
Closed in favor of soon-to-be-released variant display widget.

Issue #73, opened 2015-04-22
Top Exomiser results should be limited on view access (e.g. Public cases)


buske:
Currently, top Exomiser results are visible to all users on Public cases. This is fine and probably great, but it should be somehow limited so the whole Exomiser-processed VCF file isn't publicly available for Public cases (right now, you can just set the maximum display to ~500 genes and get the whole file). A server-side imposed maximum of ~5-10 variants should be imposed in cases where the user only has view access level.

mjshepherd:
Currently the UI uses a select input which allows all users to select values from 1 to 20. However a server-side limit is not imposed which, I agree, should be. Should users with edit rights be given the UI option to select any number of genes?

mjshepherd:
OK now without edit access the maximum number of genes that can be displayed is 10. With edit access the maximum is 500. Let me know if thats alright or if you think it should be changed.

buske:
That's alright! Thanks!

PR #74, opened 2015-04-23
Issue #73: Server-side limit imposed on number of genes in variant displ...

mjshepherd:
...ay.

### DIFF
--- a/similarity-data-impl/src/main/java/org/phenotips/data/similarity/script/ExomiserViewScriptService.java
+++ b/similarity-data-impl/src/main/java/org/phenotips/data/similarity/script/ExomiserViewScriptService.java
@@ -20,6 +20,8 @@
 package org.phenotips.data.similarity.script;
 
 import org.phenotips.data.Patient;
+import org.phenotips.data.permissions.AccessLevel;
+import org.phenotips.data.permissions.PermissionsManager;
 import org.phenotips.data.similarity.Genotype;
 import org.phenotips.data.similarity.Variant;
 import org.phenotips.data.similarity.internal.PatientGenotype;
@@ -33,6 +35,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 
+import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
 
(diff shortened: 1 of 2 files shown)


ui/src/main/resources/PhenoTips/VariantDisplay.xml
```diff
@@ -167,14 +167,15 @@
             this.optionsButton.update(new Element('a', {'class' : 'button', 'href' : '#'}).insert(this.Strings["hideOptionsButton"])).stopObserving('click', this.showOptions).observe('click', this.hideOptions.bind(this));
             
             var numberOfGenesInput = new Element('select', {'id' : 'variant-display-number-of-genes'});
-            for (var i = 1; i &lt;= 20; i++) {
+            var maxGenes = ("$xwiki.hasAccessLevel('edit')" == "true") ? 500 : 10;
+            for (var i = 1; i &lt;= maxGenes; i++) {
```
sdumitriu: Having 500 options is not good. Nobody wants exactly 432 genes, use an increasing step: `1 2 3 4 5 10 20 ... 90 100 150 200 300 500`.

similarity-data-impl/src/main/java/org/phenotips/data/similarity/script/ExomiserViewScriptService.java
```diff
@@ -51,6 +54,16 @@
 @Singleton
 public class ExomiserViewScriptService implements ScriptService
 {
+    private static final int MAXIMUM_NUMBER_OF_GENES = 10;
```
sdumitriu: The real maximum is 500, this is just a restricted maximum for unprivileged users. Use a better name.

ui/src/main/resources/PhenoTips/VariantDisplay.xml
```diff
@@ -167,14 +167,15 @@
             this.optionsButton.update(new Element('a', {'class' : 'button', 'href' : '#'}).insert(this.Strings["hideOptionsButton"])).stopObserving('click', this.showOptions).observe('click', this.hideOptions.bind(this));
             
             var numberOfGenesInput = new Element('select', {'id' : 'variant-display-number-of-genes'});
-            for (var i = 1; i &lt;= 20; i++) {
+            var maxGenes = ("$xwiki.hasAccessLevel('edit')" == "true") ? 500 : 10;
```
sdumitriu: This is wrong for two reasons.

1. This checks edit access on the `PhenoTips.VariantDisplay` document, which will be false for everybody except admins.
2. The JS file is going to be cached, so the access level that happened to be available when first requesting the JS will be used until a cache refresh happens.

Issue #76, opened 2015-05-13
Change references to the "ontology" APIs to the new "vocabulary"


mjshepherd:
Caused by https://github.com/phenotips/phenotips/issues/1659

PR #78, opened 2015-05-14
Issue #76: Change references from "ontology" to "vocabulary"

mjshepherd:


similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/DefaultPatientGenotype.java
```diff
@@ -72,7 +72,7 @@ public DefaultPatientGenotype(Patient patient)
         if (exomeManager == null) {
             ComponentManager componentManager = ComponentManagerRegistry.getContextComponentManager();
             try {
-                exomeManager = componentManager.getInstance(ExomeManager.class);
+                exomeManager = componentManager.getInstance(ExomeManager.class, "exomiser");
```
sdumitriu: This is a different issue, but good catch.

Issue #84, opened 2015-05-27
Update UI to use segmented bar


mjshepherd:

buske:
Resolved by d623306ffc93efcaa889d283399cbc6c1e819048

PR #85, opened 2015-05-29
Issue #84: Update UI to use segmented bar

mjshepherd:


ui/src/main/resources/PhenoTips/SimilarCases.xml
```diff
@@ -1558,7 +1530,6 @@
 }
 .match-details .variants-data .segmented-bar .segmented-unit-fill {
```
sdumitriu: Shouldn't all the `.segmented*` rules be removed? The segmented bar widget is supposed to be fully usable without any other style rules except setting the font size and color.

mjshepherd: The widget is usable without external css but css may still be applied to it. Float for example could either be applied in javascript or in css.

buske:
@mjshepherd please rebase

buske:
Incorporated into d623306ffc93efcaa889d283399cbc6c1e819048.

Issue #86, opened 2015-07-16
Update genetic matching to use correct solved/candidate genes


buske:
Gene sections will be unified by PR https://github.com/phenotips/phenotips/pull/1742, but this uses the same field for candidate, solved, and excluded genes, so matching will then incorrectly be done on all of them.

For matching:
- if any genes with "solved" status are present, they should be used exclusively
- elif any genes with "candidate" status are present, they should be used exclusively
- else: no genes should be used for boosting the matching.

buske:
Changes will have to be made to: https://github.com/phenotips/patient-network/blob/master/similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/DefaultPatientGenotype.java#L98-L122

buske:
Should also include gene names from candidate variants that have been entered, with pathogenic status 3-5 (VUS or higher).

buske:
Replaced by: https://phenotips.atlassian.net/browse/PN-87

PR #91, opened 2015-09-09
PN-96: Use candidate genes and variants in patient matching (Issue #86)

veronikaslc:
Fixed.

### DIFF
--- a/similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/DefaultPatientGenotype.java
+++ b/similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/DefaultPatientGenotype.java
@@ -29,6 +29,7 @@ import org.xwiki.component.manager.ComponentLookupException;
 import org.xwiki.component.manager.ComponentManager;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/DefaultPatientGenotype.java
```diff
@@ -48,9 +49,14 @@
  */
 public class DefaultPatientGenotype extends AbstractExome implements PatientGenotype
 {
+
     /** Factory for loading exome data. */
     protected static ExomeManager exomeManager;
 
+    /** List of genes variant interpretations of pathogenic status 3-5 (VUS or higher). */
+    private static final String[] TOP_INTERPRETATIONS =
+    {"Pathogenic", "Likely Pathogenic", "Variant of Unknown Significance"};
```
sdumitriu: These are translated strings, not the internal keys stored in the actual object. Which means that if PhenoTips is set in French, the value you get will be different. Why isn't the internal key used?

veronikaslc: getData("variants") returns already translated values, as I understand, getData() calls all controllers lo load() data, which comes translated for external use/display/export

PR #94, opened 2015-10-28
PN-39: return name of owner instead of referrer

buske:
This solution is unsatisfying for several reasons including:
- breaking the abstraction of the controllers
- only providing access to the owner name, not the Owner object

More "complete" solutions involving reworking AccessType ended
up resulting in more problems. Access to the PatientAccess
object appears to need to be restricted for security, and passing
it into the constructor causes problems for convert().

Feedback most welcome. @sdumitriu?

### DIFF
--- a/similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/AbstractPatientSimilarityView.java
+++ b/similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/AbstractPatientSimilarityView.java
@@ -135,10 +135,19 @@ public abstract class AbstractPatientSimilarityView implements PatientSimilarity
         // TODO: Update once there is a convenient method for accessing the owner of the patient record.
         // In the meantime, we can access the data we need through the serialized controller data.
         PatientData<String> data = this.match.getData("contact");
-        if (data == null || !data.isNamed()) {
-            return null;
+        String contact = null;
+
+        if (data != null && data.isNamed()) {
+            contact = data.get("name");
         }
-        return data.get("name");
+        // Fall back on reporter
+        if (contact == null) {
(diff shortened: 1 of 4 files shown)


buske:
@veronikaslc, do you have any thoughts or know of a better way to architect this?

sdumitriu:
Looks good to me so far, except that I would still fall back to `getReporter` if the contact controller is missing or if the `name` inside is empty.

veronikaslc:
+1

buske:
@sdumitriu: I updated it according to your suggestion.

similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/AbstractPatientSimilarityView.java
```diff
@@ -130,6 +130,27 @@ public DocumentReference getReporter()
     }
 
     @Override
+    public String getOwnerName()
+    {
+        // TODO: Update once there is a convenient method for accessing the owner of the patient record.
```
sdumitriu: I doubt there will be a more convenient method of accessing the owner, since having an owner is and will be an optional feature. Many PhenoTips instances don't have owners and access rights at all, and I think we'll even switch the default that way, since it's easier to install the access rights module when needed than to remove it.

buske: Hmm, I was expecting us to have a way to access additional patient information (e.g. Owner) in a more privacy conscious way. Right now, you have to be careful about having references to PatientAccess objects lying around since they expose the ability to set the Owner as well. The only permissions-aware interface is a ScriptService.

PR #98, opened 2015-11-06
PN-24: Better seeding of similar patients by matching the full HPO hierarchies of the reference patient's symptoms to the index.

veronikaslc:
Fixed.

### DIFF
--- a/similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
+++ b/similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
@@ -25,6 +25,8 @@ import org.phenotips.data.similarity.PatientSimilarityView;
 import org.phenotips.data.similarity.PatientSimilarityViewFactory;
 import org.phenotips.similarity.SimilarPatientsFinder;
 import org.phenotips.vocabulary.SolrCoreContainerHandler;
+import org.phenotips.vocabulary.Vocabulary;
+import org.phenotips.vocabulary.VocabularyTerm;
 
 import org.xwiki.component.annotation.Component;
 import org.xwiki.component.phase.Initializable;
@@ -32,9 +34,11 @@ import org.xwiki.component.phase.InitializationException;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Set;
 
 import javax.inject.Inject;
 import javax.inject.Named;
@@ -81,6 +85,10 @@ public class SolrSimilarPatientsFinder implements SimilarPatientsFinder, Initial
     private PatientSimilarityViewFactory factory;
 
     @Inject
+    @Named("hpo")
+    private Vocabulary ontologyService;
+
+    @Inject
     private SolrCoreContainerHandler cores;
 
     /** The Solr server instance used. */
@@ -115,6 +123,9 @@ public class SolrSimilarPatientsFinder implements SimilarPatientsFinder, Initial
     {
         //logger.error("Searching for patients using access level: {}", this.accessLevelThreshold.getName());
         SolrQuery query = generateQuery(referencePatient, prototypes);
+        if (query == null) {
+            return Collections.emptyList();
+        }
         SolrDocumentList docs = search(query);
         List<PatientSimilarityView> results = new ArrayList<PatientSimilarityView>(docs.size());
         for (SolrDocument doc : docs) {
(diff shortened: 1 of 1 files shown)


similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
```diff
@@ -155,20 +166,19 @@ private SolrQuery generateQuery(Patient referencePatient, boolean prototypes)
...
         // Ignore the reference patient itself (unless reference patient is a temporary in-memory only
         // patient, e.g. a RemoteMatchingPatient created from remote patient data obtained via remote-matching API)
         if (referencePatient.getDocument() != null) {
-            q.append("-document:" + ClientUtils.escapeQueryChars(referencePatient.getDocument().toString()));
+            q.append(" -document:" + ClientUtils.escapeQueryChars(referencePatient.getDocument().toString()));
         }
         q.append(prototypes ? " +" : " -").append("document:xwiki\\:data.MIM*");
         query.add(CommonParams.Q, q.toString());
-        //logger.error("SOLRQUERY generated: {}", query.toString());
+        logger.error("SOLRQUERY generated: {}", query.toString());
```
buske: Shouldn't be uncommented.

similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
```diff
@@ -155,16 +166,15 @@ private SolrQuery generateQuery(Patient referencePatient, boolean prototypes)
...
         StringBuilder q = new StringBuilder();
-        // FIXME This is a very basic implementation, to be revisited
-        for (Feature phenotype : referencePatient.getFeatures()) {
-            if (StringUtils.isNotBlank(phenotype.getId())) {
-                q.append(phenotype.getType() + ":" + ClientUtils.escapeQueryChars(phenotype.getId()) + " ");
-            }
+        // Whitespace-delimiter terms querying the extended phenotype field
+        Collection<String> termIds = getPresentPhenotypeTerms(referencePatient);
+        if (!termIds.isEmpty()) {
+            q.append(" extended_phenotype: " + getQueryFromTerms(termIds));
```
sdumitriu: If you leave a space between `:` and ` (`, then that is no longer a field specifier, so the query will actually look for the text `extended_phenotype` and each individual HP code in the fields specified in the `qf` parameter.

sdumitriu: Hm, it actually depends on the query type. A plain lucene query does work like this, but not a dismax query.

similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
```diff
@@ -204,4 +214,36 @@ private long count(SolrQuery query)
         }
         return 0;
     }
+
+    private Collection<String> getPresentPhenotypeTerms(Patient patient)
+    {
+        Collection<String> termIds = new ArrayList<String>();
```
sdumitriu: A list means that terms near the root of the HPO will be included many times in the query. Shouldn't a `Set` be used instead?

sdumitriu: @buske Q for you.

sdumitriu: I tested it, and specifying a term multiple times favors those terms instead of the ones appearing only once. It's boosting the more generic terms, while we actually want to emphasize the more specific ones further down the ontology.

allasm: Hmm, then why did you merge this if this needs to be fixed first?

sdumitriu: I fixed it myself, see 5401f3ee3bbcd1e3adf19ac0a526014c392b5f0c

allasm: Looks good, +1. It was not obvious from looking at https://github.com/phenotips/patient-network/pull/98#event-468759305. (BTW, in general, if merge is not urgent maybe we should wait until the author does the change just so that everyone is aware/can comment more/etc)

similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
buske: Good to know. I didn't realize it was sensitive to that space and it seemed to work. Thanks for clarifying.

similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
buske: @sdumitriu this was an intentional decision to include all ancestors multiple times to more closely mirror both the final scoring metric and the way patient phenotypes are indexed. A formal evaluation of which produces better results has not been done, but weighting ancestors more highly is actually good.

sdumitriu: My intuition tells me that boosting ancestors will actually make the matches worse. I'll try it out both ways to see how it behaves.

allasm: Just curious if we have a good metric to measure the quality of matches? From what Orion was saying today @ standup, doctors seem to prefer not the matches we theoretically think are good?

PR #99, opened 2015-11-06
PN-90: Genotypic seeding results need to be combined with phenotypic seeding.

veronikaslc:
Fixed.

### DIFF
--- a/similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/AbstractPatientGenotypeSimilarityView.java
+++ b/similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/AbstractPatientGenotypeSimilarityView.java
@@ -80,7 +80,7 @@ public abstract class AbstractPatientGenotypeSimilarityView extends AbstractExom
     }
 
     /**
-     * Return the candidate genes shared by both patients in the view.
+     * Return the set of genes listed as a candidate in either patient in the view.
      *
      * @return {@inheritDoc}
      * @see org.phenotips.data.similarity.PatientGenotype#getCandidateGenes()
@@ -93,7 +93,7 @@ public abstract class AbstractPatientGenotypeSimilarityView extends AbstractExom
             Set<String> matchGenes = this.matchGenotype.getCandidateGenes();
             Set<String> refGenes = this.refGenotype.getCandidateGenes();
             sharedGenes.addAll(matchGenes);
-            sharedGenes.retainAll(refGenes);
+            sharedGenes.addAll(refGenes);
         }
         return sharedGenes;
     }
--- a/similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/AbstractPatientGenotypeSimilarityView.java
+++ b/similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/AbstractPatientGenotypeSimilarityView.java
@@ -24,7 +24,6 @@ import org.phenotips.data.similarity.PatientGenotypeSimilarityView;
 import org.phenotips.data.similarity.Variant;
 
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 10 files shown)


similarity-data-impl/src/test/java/org/phenotips/data/similarity/internal/RestrictedPatientGenotypeSimilarityViewTest.java
```diff
@@ -457,25 +455,25 @@ public void testNoVariantsInCandidateGene()
         Assert.assertEquals(2, results.size());
 
         JSONObject top = results.getJSONObject(0);
-        Assert.assertTrue(top.getString("gene").equals("SRCAP"));
-        Assert.assertTrue(top.getDouble("score") > 0.9);
+        Assert.assertTrue(top.getString("gene").equals("NOTCH2"));
+        Assert.assertTrue(top.getDouble("score") > 0.5);
 
         // Ensure match shows underlying variant details
```
buske: Comment isn't accurate?

similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/DefaultPatientGenotypeSimilarityView.java
```diff
@@ -105,7 +106,8 @@ private PatientGenotypeManager getGenotypeManager()
      */
     private void matchGenes()
     {
-        for (String gene : getGenes()) {
+        Set<String> genes = getGenes();
```
sdumitriu: Why this change?

veronikaslc: I guess, unnecessary one

similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
```diff
@@ -204,4 +211,41 @@ private long count(SolrQuery query)
...
+                // Treat empty status as candidate
+                if (StringUtils.isBlank(status) || "solved".equals(status) || "candidate".equals(status)) {
+                    genesToSearch.add(geneName);
+                }
+            }
+        }
+        if (!genesToSearch.isEmpty()) {
+            String geneQuery = getQueryFromTerms(genesToSearch);
+            q.append(" solved_genes: " + geneQuery);
+            q.append(" candidate_genes: " + geneQuery);
```
sdumitriu: If you leave a space between `:` and `(`, then that is no longer a field specifier, so the query will actually look for the text `solved_genes` and each individual gene name in the fields specified in the qf parameter.

similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/AbstractPatientGenotypeSimilarityView.java
```diff
@@ -93,7 +93,7 @@ public boolean hasGenotypeData()
             Set<String> matchGenes = this.matchGenotype.getCandidateGenes();
             Set<String> refGenes = this.refGenotype.getCandidateGenes();
             sharedGenes.addAll(matchGenes);
-            sharedGenes.retainAll(refGenes);
+            sharedGenes.addAll(refGenes);
```
sdumitriu: This looks wrong. `sharedGenes` means genes that both patients have in common, not the union of their genes.

sdumitriu: If the meaning of the method changes, then its name should change as well, but that would be an API breakage.

buske: It was an intentional API breakage.

similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/DefaultPatientGenotypeSimilarityView.java
buske: Indeed, likely a relic from my debugging. Sorry.

PR #104, opened 2016-02-04
Update Similar Cases UIExtension to use updated consents

buske:
@allasm and @marta- please take a quick look and verify this 1-line fix is reasonable


allasm:
Looks good, +1

buske:
Thanks!

PR #105, opened 2016-02-05
PN-104: added hard limit to number of similar cases returned

buske:
1. Added hard limit to number of cases returned
1. Used PriorityQ to take them without fully sorting the list


buske:
@veronikaslc, @allasm please review

veronikaslc:
+1

PR #106, opened 2016-02-05
PN-82: reveal disorder for matchable cases

buske:
- RestrictedDisorderSimilarityView now extends DefaultDisorderSimilarityView

### DIFF
--- a/similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/RestrictedDisorderSimilarityView.java
+++ b/similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/RestrictedDisorderSimilarityView.java
@@ -21,26 +21,19 @@ import org.phenotips.data.Disorder;
 import org.phenotips.data.similarity.AccessType;
 import org.phenotips.data.similarity.DisorderSimilarityView;
 
-import org.apache.commons.lang3.StringUtils;
 import org.json.JSONObject;
 
 /**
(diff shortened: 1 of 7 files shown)


similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/RestrictedDisorderSimilarityView.java
```diff
@@ -52,73 +45,31 @@
...
-        this.reference = reference;
+        super(match, reference);
         this.access = access;
     }
 
     @Override
     public String getId()
     {
-        return this.access.isOpenAccess() && this.match != null ? this.match.getId() : null;
+        return this.access.isPrivateAccess() ? null : super.getId();
```
sdumitriu: Do we want to disclose the diagnosis, or say "Undisclosed"?

sdumitriu:
Other than the question of disclosed vs undisclosed, this looks good.

similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/RestrictedDisorderSimilarityView.java
buske: Yep, I think we want to go ahead and disclose the diagnosis for matchable cases.

buske:
Thanks for taking a look, @sdumitriu.

PR #108, opened 2016-03-18
SimilarCases remote matching UI tweaks

buske:


buske:
@marta- I teased it apart from the variant and remote-matching changes that have happened on master since then. Want to take a quick look at this to make sure it's reasonable? I left off the re-indentation for another time.

PR #109, opened 2016-04-25
Fixed gene matches background color

marta-:
Gene rows following a gene without variants were on transparent background instead of dark gray.


buske:
:+1:

PR #110, opened 2016-05-06
PN-133

itaiGershtansky:
Added methods.

### DIFF
--- a/similarity-data-api/src/main/java/org/phenotips/data/similarity/PatientSimilarityView.java
+++ b/similarity-data-api/src/main/java/org/phenotips/data/similarity/PatientSimilarityView.java
@@ -70,4 +70,20 @@ public interface PatientSimilarityView extends Patient
      *         match, with {@code 0} for patients with no similarities
      */
     double getScore();
+
+    /**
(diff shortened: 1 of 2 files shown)


itaiGershtansky:
@buske, please review

buske:
+1

similarity-data-api/src/main/java/org/phenotips/data/similarity/PatientSimilarityView.java
```diff
@@ -70,4 +70,20 @@
      *         match, with {@code 0} for patients with no similarities
      */
     double getScore();
+
+    /**
+     * Get the phenotypic similarity score for this patient match.
+     *
+     * @return the similarity score, between 0 (a poor match) and 1 (a good
```
sdumitriu: @buske I see that the Javadoc for `getScore()` (see above) uses `[-1, 1]` for the score range. Is that valid or deprecated?

buske: Scores are never negative, so we could consider it deprecated. I think the concept of an "opposite" patient description is rather tenuous.

PR #112, opened 2016-05-24
PN-142: Limit number of matches that appear in the UI

buske:
This is a temporary fix until a "Show more results" solution is implemented.

### DIFF
--- a/ui/src/main/resources/PhenoTips/SimilarCases.xml
+++ b/ui/src/main/resources/PhenoTips/SimilarCases.xml
@@ -205,6 +205,7 @@
     _NOMATCH_MARKER : "no-match",
     _NOMATCH_TEXT : "No match",
     _DEFAULT_REFRESH_BUTTON_TEXT : "Refresh Matches",
+    _MAX_RESULTS_DISPLAYED : 10,
     DISEASE_DB_LINKS : {
       "MIM:" : "http://www.omim.org/entry/"
     },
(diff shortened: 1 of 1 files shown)


buske:
@allasm and @veronikaslc please give this a quick review

buske:
Once this PR is merged, we'll hotfix PC and add the Monarch MME endpoint to the UI.

ui/src/main/resources/PhenoTips/SimilarCases.xml
```diff
@@ -318,6 +319,8 @@
       }
       data.featuresCount = data.query.features ? data.query.features.size() : 0;
 
+      data.resultsCount = Math.min(data.resultsCount, this._MAX_RESULTS_DISPLAYED);
+      data.results = data.results.slice(0, data.resultsCount);
```
allasm: I quickly browsed through the MME code and this file, and don't see where we sort the results by score? They may happen to be sorted by our server (again, MME code does not sort, but I think the underlying findSimilarCases() in patient-network does) - but is there a guarantee remote servers will return results in the sorted order?

buske: Good point. No, there's no guarantee. We should sort the results before paginating.

sdumitriu:
@buske Any news on this?

buske:
@sdumitriu MME servers are not guaranteed to return their results in sorted order. However, all currently do. As a result, we hotfixed PC with this fix and enabled matching with Monarch. This solution is okay for now, but it would be best to sort by score first.

PR #115, opened 2016-10-12
PC-132: Upgrade to Exomiser 7.2.2

buske:
- Updates cron script and Exomiser settings template to be 7.2.2-compatible

### DIFF
--- a/scripts/exomiser_cron.py
+++ b/scripts/exomiser_cron.py
@@ -42,7 +42,8 @@ MODES_OF_INHERITANCE = {
     'HP:0001419': 'X',
     'HP:0001423': 'X',
     }
-
+DEFAULT_EXOMISER_TEMPLATE = os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                                         'exomiser_settings.txt')
 
 class RecordLockedException(Exception):
     pass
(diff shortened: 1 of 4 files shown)


ui/src/main/resources/PhenoTips/MatchContact.xml
```diff
@@ -63,7 +63,7 @@
   #set ($emailObj = $doc.getObject('XWiki.Mail'))
```
sdumitriu: That should be `PN-132`.

scripts/exomiser_cron.py
```diff
@@ -42,16 +42,16 @@
     'HP:0001419': 'X',
```
sdumitriu: Use the issue ID in the commit message.

scripts/exomiser_cron.py
```diff
@@ -230,7 +230,7 @@ def parse_phenotypes(data):
     # Add in prenatal phenotype
     features.extend(data.get('prenatal_perinatal_phenotype', {}).get('prenatal_phenotype', []))
     for feature in features:
-        if feature.get('type') == 'phenotype' and feature.get('observed') == 'yes':
+        if feature.get('observed') == 'yes':
```
sdumitriu: Does this mean that now prenatal phenotypes are taken into account by the exomiser? Isn't that an improvement that also requires an issue?

PR #116, opened 2016-11-11
PN-150: MatchContact breaks if case owner is a group

itaiGershtansky:


itaiGershtansky:
Together with https://github.com/phenotips/phenotips/pull/2151

itaiGershtansky:
@buske please review

PR #122, opened 2016-11-22
VS-4, PN-110: API changes to support variant-store implementation

buske:
This updates patient network APIs to accommodate the variant store, but should be backwards compatible. To test.


similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/GAVariantVariant.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.commons.lang.StringUtils;
+import org.ga4gh.GACall;
+import org.ga4gh.GAVariant;
+
+/**
+ * A variant from the variant store. Annotated by Exomiser.
+ *
+ * @version $Id$
+ */
+public class GAVariantVariant extends AbstractVariant
```
buske: Rename to VariantStoreVariant?

similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/GAVariantVariant.java
```diff
@@ -0,0 +1,63 @@
...
+
+        setGenotype(gaVariant.getReferenceBases(),
+                StringUtils.join(gaVariant.getAlternateBases(), ','),
+                StringUtils.join(genotype, '/'));
+
+        setEffect(VariantUtils.getInfo(gaVariant, GAVariantInfoFields.GENE_EFFECT));
+
+        Object value = VariantUtils.getInfo(call, GACallInfoFields.EXOMISER_VARIANT_SCORE);
+        if (value == null || value.equals("null")) {
+            setScore(0D);
```
buske: Should setScore(null)

similarity-data-impl/src/main/java/org/phenotips/data/similarity/script/ExomiserViewScriptService.java
```diff
@@ -89,36 +89,56 @@ public boolean hasGenotype(Patient patient)
      */
     public JSONArray getTopGenesAsJSON(Patient patient, int g, int v)
     {
-        JSONArray result = new JSONArray();
-        if (patient == null || g <= 0) {
-            return result;
+        JSONArray variantsJSON = new JSONArray();
+
+        if (patient == null || g < 0) {
```
buske: Unless we want 0 to have the special meaning of "all", this should be kept as `g <= 0`

PR #128, opened 2016-12-15
PN-162: Stop matching on manual variants

buske:
This PR removes the matchmaking behavior of candidate variants, instead only using the candidate gene section. This behavior is better, until such a time as we allow more explicit control over which variants and genes are used for matching.


buske:
I should have added the new tests directly and then refactored later, but I wanted to refactor the code to make the new tests easier to express. There were no tests for manual variants, nor rejected genes.

PR #131, opened 2017-02-03
PN-161: Notifications does not handle contacts for remote matches correctly; also PN-164; PN-165; PN-166; PN-167; PN-169; PN-170; PN-171; PN-173; PN-178

veronikaslc:
To test PN-161 build in order:
- `phenotips` branch `rm-refactor` on top of milestone-5 with -Pquick
- `patient-network` branch `PN-161` with -Pquick
- `remote-matching` branch `PN-161`
- `phenomecentral`
Create a group with email info. Create and approve new user with email info. Create a new patient as admin, group and as a new user with same candidate gene and any phenotype. Check that `Contact` column for `Similar cases` section is filled in correctly.
Check that matching notification table has contact info filled in correctly.

### DIFF
--- a/matching-notification-ui/src/main/resources/MatchingNotification/PatientMatchNotificationEmailTemplate.xml
+++ b/matching-notification-ui/src/main/resources/MatchingNotification/PatientMatchNotificationEmailTemplate.xml
@@ -98,6 +98,9 @@
 #set($space-between-matches = 'margin-top : 25px;')
 #set($disclaimer = 'font-size : 10px;')
 #set($patient-id = 'white-space : no-wrap;')
+#if ("$!subjectPatient.serverId" != "")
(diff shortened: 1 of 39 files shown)


veronikaslc:
compiles with -Pquick option

matching-notification-ui/src/main/resources/MatchingNotification/PatientMatchNotificationEmailTemplate.xml
```diff
@@ -98,6 +98,9 @@
 #set($space-between-matches = 'margin-top : 25px;')
 #set($disclaimer = 'font-size : 10px;')
 #set($patient-id = 'white-space : no-wrap;')
+#if ("$!subjectPatient.serverId" != "")
```
sdumitriu: Shouldn't this be `#if (!`, meaning that the patient shouldn't have a server ID, since it's supposed to be local?

PR #132, opened 2017-02-18
PN-181: Add option to send matching notification to one user only

veronikaslc:


veronikaslc:
Strange thing, may be pushed wrong version of code. Closing this PR, branch name is wrong anyways.

PR #134, opened 2017-04-01
PN-183: Gene ID is shown instead of gene symbol in Similar patients -> Gene matching breakdown

veronikaslc:


ui/src/main/resources/PhenoTips/SimilarCases.xml
```diff
@@ -54,7 +54,17 @@
   "resultsCount" : $matches.size(),
   "results" : [
 #foreach ($p in $matches)
-  $p.toJSON()#if ($foreach.hasNext()),
+   #set ($match = $p.toJSON())
```
sdumitriu: Wouldn't it be simpler to do this [in java](https://github.com/phenotips/patient-network/blob/master/similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/DefaultPatientGenotypeSimilarityView.java#L255)?

veronikaslc: I was actually thinking it would be better to use Gene class from PT-3072 PR. But it involves refactoring gene code in patient-network (which was planned to be discussed), so I can just quick fix it in Java for now if urgently needed.

veronikaslc: @sdumitriu comments addressed

PR #139, opened 2017-05-02
PN-189: In the Similar patients UI, the disorder name is missing its …

marta-:
…first word


sdumitriu:
I wonder if `name_translated` works.

marta-:
We're not returning the translated name in this json. I'll open a separate task for it.

PR #138, opened 2017-05-01
PN-188: Improve variant-store-related UI

katpavlova:
Much of this commit was originally authored by @meatcar and @buske .

- Make variant-store-related UI strings neatly translatable
- Some minor UI improvements to the patient-matching section of the PhenomeCentral patient form (view-mode.)

PR #140, opened 2017-05-04
PN-191: Add an indicator while searching for matches

veronikaslc:


sdumitriu:
@katpavlova also introduced translations in #138, was this work done twice?

veronikaslc:
@sdumitriu removed unnecessary commit

PR #141, opened 2017-05-15
PN-192: Allow pagination of the similarity results

veronikaslc:


similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
```diff
@@ -72,9 +72,6 @@
     /** The number of records to fully score. */
     private static final int SEED_QUERY_SIZE = 50;
 
-    /** The number of records to return after full scoring. */
-    private static final int MAX_RESPONSE_SIZE = 10;
```
allasm: did you test it on a large database, e.g. PhenomeCentral? Are you sure pagination only in JS (not server-side) works fast enough, e.g. if there are 5000 matches wont it be too slow to even transfer them to the client browser?

PR #143, opened 2017-05-18
[misc] Fix for disappearing phenotypes bug in matching notification email

veronikaslc:


matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPhenotypesMap.java
```diff
@@ -219,17 +219,29 @@ public static void reorder(List<Map<String, String>> predefined1, List<Map<Strin
             if (StringUtils.equals(name1, name2)) {
                 common1.add(item1);
```
allasm: just for the sake of future developers, i would extend the method comment and mention that the incoming lists (`predefined1` and `predefined2`) are supposed to be sorted, as otherwise the algorithm does not work

allasm: we also don't need `if (next1)` and `if (next2)` checks, they are equivalent to `item1 != null` and `item2 != null` checks

allasm: @sdumitriu if you don't like checkstyle excludes merge in the per-method checkstyle exclusions branch, which will make it much cleaner and less-excluded :)

sdumitriu: It's not the excludes file that bothers me, but that the exclusions are needed.

allasm: I understand...  I wonder if removing `next1` and `next2` and combining the `if` and `while` after the loop may decrease the complexity enough to avoid excllusions

(e.g. instead of separate if and while have something like
```while (item1 != null) {
              unique1.add(item1);
              item1 = iterator1.hasNext()? iterator1.next() : null;
    }
```

veronikaslc: @allasm if we change `if (next1)` to `item1 != null` the cycle wont start, cause `item1` is `null` in the beginning.

allasm: this is for the if statement on line 237 (in the new code) and cycle on line 245 - if `item1` is `null` at that point, the cycle shouldn't start

veronikaslc: @allasm @sdumitriu  I made an alternative implementation of this reordering function in the next commit to avoid exclusions and make process more simple & clear.

PR #146, opened 2017-05-26
PN-192: Allow pagination of the similarity results

a-gvo:
Back-end component completed.

### DIFF
--- a/ui/src/main/resources/PhenoTips/SimilarCases.xml
+++ b/ui/src/main/resources/PhenoTips/SimilarCases.xml
@@ -244,11 +244,38 @@
           this._disclaimerDialog = new PhenoTips.widgets.ModalPopup(this._disclaimerContainer, false, {'title': "$escapetool.javascript($services.localization.render('phenotips.similarCases.disclaimerDialogTitle'))", 'removeOnClose': false, 'resetPositionOnShow': false});
         }
         this._omimPrototypeMode = ((options.omimSection || "") != "");
+        if (options.enablePagination) {
+          this._PAGE_COUNT_TEMPLATE = "Showing similar cases __number_range__ out of ___total___ per page of ";
+          this.page = 1;
+          this.prevPage = $('pagination-prev');
+          this.nextPage = $('pagination-next');
+          this.paginationContent = $('pagination-content');
+          this.attachNavigationListeners();
+        }
       }
       this.launchSearchExisting();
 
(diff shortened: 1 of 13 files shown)


ui/src/main/resources/PhenoTips/SimilarCases.xml
```diff
@@ -347,6 +439,63 @@
...
+        var _this = this;
+        pageSpan.observe("click", function(ev){
+            ev.stop();
+            _this.page = ev.element().innerHTML;
+            _this.launchSearchExisting();
+        });
+        return pageSpan;
+    },
+
+    generatePaginationPageLinks : function() {
```
veronikaslc: the question from @allasm that he originally asked in my first implementation: if `this.totalPages` is 1000+, will all page links be displayed at once?

allasm: I think the problem with page links is not a big deal, since having 1000+ patients is likely, 100+ page sis not, and iven if they have 1000+ pages i don't think we want to page pages :)

a-gvo: Fixed.

sashaandjic:
I'm unable to test it.
screen shot 2017-06-14 at 11 34 43 pm

sashaandjic:
tested. works as expected.

ui/src/main/resources/PhenoTips/SimilarCases.xml
```diff
@@ -198,7 +198,7 @@
       <code>var PhenoTips = (function (PhenoTips) {
   var widgets = PhenoTips.widgets = PhenoTips.widgets || {};
   widgets.Matcher = Class.create({
-    matchServiceURL : "${xwiki.getURL('PhenoTips.SimilarCases', 'get')}",
+    matchServiceURL : XWiki.contextPath + "/rest/patients/___patient_id___/similar-cases",
```
veronikaslc: @a-gvo with this change lines 40-67 becoming redundant?

a-gvo: I believe you're right.

veronikaslc:
@a-gvo One more thing. Due to the most recent fixes of PN-214 we now do not check permissions to show the email address or initiate contact. So things like `processEmail()` and `isManager` are not needed anymore. Hopefully this will be an easy last fix for this PR.

a-gvo:
@veronikaslc Fixed.

PR #149, opened 2017-07-11
PN-208: Use internal score for remote matches

veronikaslc:


buske:
This is a partial solution, as the remote matches will now appear to be unsorted, yes?

PR #150, opened 2017-07-11
PN-202: Improve the look of matching details

veronikaslc:


buske:
@veronikaslc Could you please include a screenshot of the updated UI?

buske:
How does this behave if the gene has variants in the two cases? Should that section have the headings as well, or will that further confuse things because then it looks like WNK1 is only in the current patient?

veronikaslc:
`course.phenomecentral.org` is updated with changes, you can test the look there

ui/src/main/resources/PhenoTips/SimilarCases.xml
```diff
@@ -684,7 +684,7 @@
 
       var row = this._getEmptyTableRow(modeOfInheritanceTable);
 
-      var handleEmptyArray = function(moiArray) {return (moiArray &amp;&amp; moiArray.length &gt; 0) ? moiArray : ['-'];};
+      var handleEmptyArray = function(moiArray) {return (moiArray &amp;&amp; moiArray.length &gt; 0) ? moiArray : ['Unknown'];};
```
sdumitriu: Doesn't this need to be translated?

veronikaslc:
@sdumitriu Added translations

PR #158, opened 2017-11-23
PN-203: Group the matched phenotypes under the broader system branches, e.g. Abnormality of the head, Abnormality of the neck

veronikaslc:
- Separated genotype similarity to own component


matching-notification-api/src/main/java/org/phenotips/matchingnotification/notification/internal/DefaultPatientMatchEmail.java
```diff
@@ -239,4 +242,43 @@ public String getSubjectPatientId()
...
+    {
+        // we reconstruct features from corresponding details saved with match
+        // because there features are already Reordered via DefaultPhenotypesMap
+        Set<? extends Feature> matchFeatures = readFeatures(match.getMatchedDetails());
+        Set<? extends Feature> refFeatures = readFeatures(match.getReferenceDetails());
+
+        // passing null in place of access because the user is admin,
+        // access not used in DefaultPatientPhenotypeSimilarityView
+        PatientPhenotypeSimilarityView featuresView =
+            new DefaultPatientPhenotypeSimilarityView(matchFeatures, refFeatures, null);
```
sdumitriu: This is an implementation detail from another module, so it should be hidden. Use a factory instead, even if all the factory does is instantiate and return this `DefaultPatientPhenotypeSimilarityView`.

similarity-data-api/src/main/java/org/phenotips/data/similarity/FeatureClusterView.java
```diff
@@ -27,13 +27,13 @@
...
 
 /**
- * View of the relationship between a collection of features in each of two patients.
+ * View of the relationship between a collection of features in each of two patients within one cluster of features.
  *
  * @version $Id$
  * @since 1.0M1
  */
 @Unstable
-public interface FeatureClusterView extends Feature
```
sdumitriu: Why is this change needed?

similarity-data-impl/src/test/java/org/phenotips/data/similarity/internal/RestrictedPatientSimilarityViewTest.java
```diff
@@ -556,7 +570,7 @@ public void testToJSONWithPublicAccess() throws ComponentLookupException
         JSONObject result = o.toJSON();
         Assert.assertEquals(o.getContactToken(), result.getString("token"));
         JSONArray clusters = result.getJSONArray("featureMatches");
-        Assert.assertTrue(clusters.length() >= 2);
+        //Assert.assertTrue(clusters.length() >= 2);
```
sdumitriu: Why?

similarity-data-rest/pom.xml
```diff
@@ -34,9 +34,9 @@
   <name>PhenoTips - Patient network - Java REST APIs for finding similar patients</name>
   <dependencies>
     <dependency>
-      <groupId>${project.groupId}</groupId>
```
sdumitriu: All the changes in this file seem unneeded, please exclude them from the PR.

similarity-data-api/src/main/java/org/phenotips/data/similarity/FeatureClusterView.java
veronikaslc: We do not need to extend Feature

matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientMatch.java
```diff
@@ -21,9 +21,10 @@
 import org.phenotips.data.Patient;
 import org.phenotips.data.PatientRepository;
 import org.phenotips.data.similarity.PatientSimilarityView;
+import org.phenotips.data.similarity.phenotype.DefaultPhenotypesMap;
```
allasm: How does\did this work, `DefaultPhenotypesMap` seems to have been used before but not imported?

matching-notification-api/src/main/java/org/phenotips/matchingnotification/notification/internal/DefaultPatientMatchEmail.java
veronikaslc: @sdumitriu addressed in the last commit, if I got the idea right (please check).

veronikaslc:
Closed, is re-implemented in the corresponding commit in `stable-merged-test` branch. To be cherry-picked to master after merge to `stable-1.1-x`.

PR #160, opened 2017-11-23
[misc] Save matches found for single patient in separate thread

veronikaslc:


similarity-data-rest/pom.xml
```diff
@@ -27,7 +27,7 @@
   </parent>
 
   <properties>
-    <coverage.instructionRatio>1.00</coverage.instructionRatio>
+    <coverage.instructionRatio>0.93</coverage.instructionRatio>
```
allasm: Maybe we should write a test for that? At least one which just checks that `this.matchingNotificationManager.saveIncomingMatches()` is called at some point

allasm:
Once code is approved this needs to be tested since changes are not obvious

PR #163, opened 2017-11-29
PN-232: Sort matches by score

veronikaslc:


allasm:
BTW, does pagination sort by score?

veronikaslc:
@allasm good point, id did not. But I gitjust made a PR :)

similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
```diff
@@ -149,36 +149,7 @@ public long countSimilarPatients(Patient referencePatient)
...
-        List<PatientSimilarityView> topResults = new ArrayList<>();
-        int docSize = pq.size();
-        for (int i = 0; i < docSize; i++) {
-            PatientSimilarityView item = pq.poll();
-            if (item != null) {
-                topResults.add(item);
-            }
-        }
-        return topResults;
+        return results;
```
allasm: There is a change in functionality now, `find()` used to return sorted results, now it does not. I know this is taken care of by the new sorting functionality in RM-pagination, but conceptually it seems better to sort here (so that all users of `find()` get sorted results) instead of higher in the code stack?

veronikaslc: Do all users of `find()` require sorted results? And will be sorting order  preserved down stack? Will be any matches filtered/excluded from list later? Seems better to do sorting for final results at the place of clusterview initialised.

veronikaslc:
@allasm changes made

allasm:
To be tested before merging

PR #164, opened 2017-11-29
PN-232: Sort matches by score in the REST similar patients finder results

veronikaslc:


similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/DefaultMatchedPatientClusterView.java
```diff
@@ -137,6 +138,18 @@ private boolean convertEmpty(final int fromIndex, final int toIndex)
      */
     private JSONArray buildMatchesJSONArray(final int fromIndex, final int toIndex)
     {
+        // Sort by score
```
allasm: this way we re-sort every time we request the next page. I think if we sort we should do it once when we initialise/instantiate this list (i.e. in constructor)

allasm:
Don't make everything `[misc]`, create an issue

similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/DefaultMatchedPatientClusterView.java
veronikaslc: As far as I understand we initialise this lidt each time we request a next page

veronikaslc: But sorting should happen in constructor, I'll relocate it.

similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/DefaultMatchedPatientClusterView.java
```diff
@@ -22,6 +22,7 @@
 import org.phenotips.data.similarity.PatientSimilarityView;
 
 import java.util.Collections;
+import java.util.Comparator;
```
allasm: Another minor fix: should be `since 1.4` not `since 1.2`. Github does not let me comment on that line, it is a few lines below this one

veronikaslc: Why 1.4?

allasm:
Code approved, to be merged after testing

veronikaslc:
closed in favour of https://github.com/phenotips/patient-network/pull/163

PR #167, opened 2017-12-05
Stable merged test

veronikaslc:


communication/src/main/java/org/phenotips/messaging/internal/DefaultActionManager.java
```diff
@@ -211,14 +210,7 @@ private String getEmail(DocumentReference userDocument)
...
     {
-        List<PatientMatch> successfulMatches = new LinkedList<>();
-        List<PatientMatch> matchesForPatient = this.matchStorageManager.loadMatchesByReferencePatientId(patientId);
-        for (PatientMatch match : matchesForPatient) {
-            if (match.getMatchedPatientId().equals(matchId)) {
-                successfulMatches.add(match);
-                break;
-            }
-        }
+        List<PatientMatch> successfulMatches = this.matchStorageManager.loadMatchesBetweenPatients(patientId, matchId);
```
allasm: There is a problem here: which servers do we expect here? I.e. is `patientId` a local ID or remote ID? Is `matchId` a local id or remote id?

veronikaslc:
commits are codereviewed

communication/src/main/java/org/phenotips/messaging/internal/DefaultActionManager.java
```diff
@@ -126,7 +125,8 @@ public int sendInitialMails(Connection connection, Map<String, Object> options)
             mail.setTextPart((String) options.get(OPTIONS_MESSAGE_FIELD));
             mail.setSubject((String) options.get(SUBJECT_FIELD_STRING));
             mailsender.sendMail(mail, context);
-            setNotified((String) options.get("patientId"), (String) options.get("matchId"));
+            setNotified((String) options.get("patientId"), this.convertServerId((String) options.get("serverId")),
+                    this.convertServerId((String) options.get("matchId")), (String) options.get("matchServerId"));
```
sdumitriu: This looks wrong to me. Shouldn't `convertServerId` apply to `matchServerId` instead of `matchId`?

PR #169, opened 2018-01-13
PN-253: Create a configuration table to show the last date+time matching was run; PN-252: Run local matches search only for patients updated/created since last search

veronikaslc:


matching-notification-api/src/main/java/org/phenotips/matchingnotification/finder/internal/LocalMatchFinder.java
```diff
@@ -53,6 +77,38 @@
...
+
+    private Date previousStartedTime;
+
+    private XWikiDocument prefsDoc;
+
+    @Override
+    public void initialize() throws InitializationException
+    {
+        XWikiContext context = this.provider.get();
+        this.previousStartedTime = null;
```
allasm: i think style checker will complain about initializing to `null` since that is already the default value

matching-notification-api/src/main/java/org/phenotips/matchingnotification/finder/MatchFinder.java
```diff
@@ -43,4 +43,13 @@
      */
     List<PatientMatch> findMatches(Patient patient);
 
+    /**
+     * Record start time for running matches search.
+     */
+    void recordStartMatchesSearch();
```
allasm: I would name this method a bit different to at least differentiate it from the same method in `MatchFinderManager.java`

E.g. this can be named `RecordMatcherStart`, and the other `RecordAllMatchesStart` or smth like that , to visually differentiate them and what they do

PR #170, opened 2018-01-16
PN-256: Non-consented patients are considered as matches for incoming MME requests

allasm:
@sashaandjic please test (I will explain how on Slack)

### DIFF
--- a/similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
+++ b/similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
@@ -17,6 +17,7 @@
  */
 package org.phenotips.similarity.internal;
 
+import org.phenotips.data.ConsentManager;
 import org.phenotips.data.Feature;
 import org.phenotips.data.Patient;
 import org.phenotips.data.PatientData;
@@ -99,6 +100,9 @@ public class SolrSimilarPatientsFinder implements SimilarPatientsFinder, Initial
     @Inject
     private SolrCoreContainerHandler cores;
 
+    @Inject
+    private ConsentManager consentManager;
+
     /** The Solr server instance used. */
     private SolrClient server;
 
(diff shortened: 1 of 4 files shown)


similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
```diff
@@ -111,13 +115,19 @@ public void initialize() throws InitializationException
     @Override
     public List<PatientSimilarityView> findSimilarPatients(Patient referencePatient)
     {
-        return find(referencePatient, false);
+        return findSimilarPatients(referencePatient, null);
```
veronikaslc: why not simply `return find(referencePatient, null, false);`

similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
```diff
@@ -111,13 +115,19 @@ public void initialize() throws InitializationException
...
+    }
+
+    @Override
+    public List<PatientSimilarityView> findSimilarPatients(Patient referencePatient, String requiredConsentId)
```
veronikaslc: Why not the list of all possible (in future) consents? We have only one just for now.

similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
```diff
@@ -144,6 +154,10 @@ public long countSimilarPatients(Patient referencePatient)
                 // Leftover patient in the index, should be removed
                 continue;
             }
+            // check consents, if required
+            if (requiredConsentId != null && !consentManager.hasConsent(matchPatient, requiredConsentId)) {
```
veronikaslc: either `check consent` or it should a be a list of consents

similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
allasm: more modular, this method is a shortcut for `findSimilarPatients` for the case when we have no consents, so w edo exactly that here. The fact that that method in turn calls something else is its internal implementation which we dont care about here, maybe in the future that will change, but this way we will only have to change it in one place (ther,e not both there and here)

similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
allasm: I thought about this, just no sure what is a good convenient way to do this (just alist may work, but is ugly on the calling side, when you have to create a list of one item all the time). We can upgrade to include a list later when we have a better idea about a use case

sdumitriu: varargs is easy.

PR #171, opened 2018-01-16
PN-239: Allow filtering matches separately by phenotypical and genotypical scores

veronikaslc:


matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/unnotifiedMatchesTable.js
```diff
@@ -77,12 +69,16 @@ var PhenoTips = (function (PhenoTips) {
     _showMatches : function()
     {
         var score = this._checkScore('show-matches-score', 'show-matches-messages');
-        if (score == undefined) {
+        var phenScore = this._checkScore('show-matches-phen-score', 'show-matches-messages');
+        var genScore = this._checkScore('show-matches-gen-score', 'show-matches-messages');
+        if (score == undefined && phenScore == undefined && genScore == undefined) {
```
allasm: after this check we know at least one of them is defined - but some may still be undefined. So I think we should set all the ones that are `undefined` to 0 after this check

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/unnotifiedMatchesTable.js
```diff
@@ -123,12 +119,11 @@ var PhenoTips = (function (PhenoTips) {
     _checkScore : function(scoreFieldName, messagesFieldName) {
         var score = this._$('#' + scoreFieldName).val();
         if (score == undefined || score == "") {
-            this._utils.showHint(messagesFieldName, "$services.localization.render('phenotips.matchingNotifications.emptyScore')");
-            return;
+            return 0;
         } else if (isNaN(score)) {
             this._utils.showHint(messagesFieldName, "$services.localization.render('phenotips.matchingNotifications.invalidScore')");
             return;
```
allasm: I think this should `return undefined` now so that this can be caught in a check above and search cancelled

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/unnotifiedMatchesTable.js
```diff
@@ -123,12 +119,11 @@ var PhenoTips = (function (PhenoTips) {
...
-        };
+        }
         scoreNumber = Number(score);
         if (scoreNumber < 0 || scoreNumber > 1) {
             this._utils.showHint(messagesFieldName, "$services.localization.render('phenotips.matchingNotifications.invalidScore')");
```
allasm: same here

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/unnotifiedMatchesTable.js
allasm: also replace `&&` with `||`

veronikaslc:
@allasm updated last commit

PR #174, opened 2018-01-24
PN-261: Add ability to select MME or local servers for matching search

veronikaslc:


matching-notification-api/src/main/java/org/phenotips/matchingnotification/script/MatchingNotificationScriptService.java
```diff
@@ -193,7 +156,7 @@ public String setStatus(String ids, String status)
      */
     public boolean saveLocalMatches(List<PatientSimilarityView> similarityViews, String patientId)
```
allasm: Not related to the pull request, but related to your changes: since you are reworking the APIs, we should also remove this method completely, matches should always be saved wherever they are found (on the back-end), velocity script code (and this method is only called form velocity) should not have to explicitly save matches

matching-notification-api/src/main/java/org/phenotips/matchingnotification/finder/MatchFinder.java
```diff
@@ -35,38 +36,25 @@
      */
     int getPriority();
 
-    /**
-     * Returns a name which can be used to select a specific matcher to be run.
-     * @return a name
-     */
-    String getName();
```
allasm: We should have a `getSupportedServerIdList()` or something like that instead, otherwise there is no way to know which matcher can find matches for which server

allasm: In fact maybe we want each matcher to register itself as a matcher handling requests to serverIDs X, Y and Z

matching-notification-api/src/main/java/org/phenotips/matchingnotification/finder/internal/LocalMatchFinder.java
```diff
@@ -54,37 +54,32 @@ public int getPriority()
     }
 
     @Override
-    public String getName()
+    public List<PatientMatch> findMatches(List<Patient> patients, Set<String> serverIds,
+        boolean onlyUpdatedAfterLastRun)
```
allasm: There should be a check that throws an exception if there is an unsupported `serverId` in the set of requested `serverIds`

matching-notification-api/src/main/java/org/phenotips/matchingnotification/finder/internal/DefaultMatchFinderManager.java
```diff
@@ -66,75 +65,16 @@
...
-            service.recordStartMatchesSearch();
-        }
-    }
-
-    private void recordFindAllEnd(Set<String> matchersToUse)
-    {
-        for (MatchFinder service : this.matchFinderProvider.get()) {
-            if (matchersToUse != null && !matchersToUse.contains(service.getName())) {
+        for (MatchFinder matchFinder : this.matchFinderProvider.get()) {
+            if (matchFinder instanceof LocalMatchFinder && !serverIds.contains("local")) {
```
allasm: This is a bad pattern (`instanceof`). Instead we should design a system to know which matcher(s) to ruun, but one that doe snot rely on class names. E.g. see my comments above and below. Not sure whats the best system yet, need to discuss

matching-notification-api/src/main/java/org/phenotips/matchingnotification/script/MatchingNotificationScriptService.java
veronikaslc: Since we use REST in master no pint to rework it now.

matching-notification-api/src/main/java/org/phenotips/matchingnotification/finder/internal/DefaultMatchFinderManager.java
veronikaslc: moved a check to the `LocalMatchFinder`

matching-notification-api/src/main/java/org/phenotips/matchingnotification/finder/internal/LocalMatchFinder.java
veronikaslc: By design each `MatchFinder` selects server Ids that is supports and search matches on each of them.

sashaandjic:
It seems to run matches only for one out of many selected servers.

veronikaslc:
rebased on `stable-1.1.7x`

PR #177, opened 2018-01-26
PN-146: Phenotypic and genotypic features should not be obfuscated

veronikaslc:


similarity-phenotype-view/pom.xml
```diff
@@ -27,7 +27,7 @@
   </parent>
 
   <properties>
-    <coverage.instructionRatio>0.15</coverage.instructionRatio>
+    <coverage.instructionRatio>0.13</coverage.instructionRatio>
```
sdumitriu: This isn't good.

veronikaslc: I did not add any code

PR #182, opened 2018-02-02
PN-270: Add search by server ID to the matching notification table

veronikaslc:


matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -31,7 +31,9 @@ define(["jquery", "dynatable"], function($, dyna)
                     || match.matched.externalId.includes(this._searchFilter)
                     || match.reference.externalId.includes(this._searchFilter)
                     || match.matched.emails.toString().includes(this._searchFilter)
-                    || match.reference.emails.toString().includes(this._searchFilter) ) 
+                    || match.reference.emails.toString().includes(this._searchFilter)
+                    || match.matched.serverId.includes(this._searchFilter)
```
sdumitriu: Why toString is not called here?

veronikaslc: because `serverId` is a string like `externalId`, not array like `emails`

sashaandjic:
tested

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
sdumitriu: Then why is it called on the next line?

PR #184, opened 2018-02-02
PN-259: "Matcher run info" document improvement

allasm:
To test: usability of "refresh matches" page; for each run there should be more data printed in a new table below the refresh matches section; for long runs it should mark the server row as "not completed"

### DIFF
--- a/matching-notification-ui/src/main/resources/MatchingNotification/MatchingRunInfoClass.xml
+++ b/matching-notification-ui/src/main/resources/MatchingNotification/MatchingRunInfoClass.xml
@@ -101,6 +101,7 @@
       <unmodifiable>0</unmodifiable>
       <validationMessage/>
       <validationRegExp/>
+      <numberType>integer</numberType>
       <classType>com.xpn.xwiki.objects.classes.NumberClass</classType>
     </numPatientsCheckedForMatches>
     <numErrors>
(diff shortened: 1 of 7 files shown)


matching-notification-ui/src/main/resources/MatchingNotification/MatchingRunInfoClass.xml
```diff
@@ -91,17 +91,53 @@
...
-      <prettyName>Number of local patients updated</prettyName>
+      <prettyName>Number of local patients processed</prettyName>
       <size>8</size>
       <unmodifiable>0</unmodifiable>
       <validationMessage/>
       <validationRegExp/>
       <classType>com.xpn.xwiki.objects.classes.NumberClass</classType>
-    </numPatientsUsedForLastMatchRun>
+    </numPatientsCheckedForMatches>
+    <numErrors>
```
sdumitriu: Don't manually edit these XMLs, this XClass is not valid since it's missing a `<numberType>` setting.

allasm: I just re-used a piece of some other XML we have in PT, whats a `numberType`? This document can be imported using theimport option without errors, so whats not valid?

PR #186, opened 2018-02-08
[misc] Cherry-picked master commits on stable-1.1.x

veronikaslc:


matching-notification-api/src/main/resources/ApplicationResources.properties
```diff
@@ -2,9 +2,7 @@ phenotips.matching.ajaxutils.requestSent=Sending request...
 phenotips.matching.ajaxutils.done=Done
 phenotips.matching.ajaxutils.failure=Request failed
 
-phenotips.matchingNotifications.emptyScore=Empty score
 phenotips.matchingNotifications.invalidScore=Score must be a number between 0 and 1
-phenotips.matchingNotifications.loadingMatches=Loading matches
```
allasm: +1, these two strings were never used

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/notifiedMatchesTable.js
```diff
@@ -79,7 +79,7 @@ var PhenoTips = (function (PhenoTips) {
             var [successfulIds, failedIds] = this._utils.getResults(ajaxResponse.responseJSON.results);
 
             if (failedIds.length > 0) {
-                alert("Sending notification failed for the matches with the following ids: " + failedIds.join());
+                alert("$escapetool.javascript($services.localization.render('phenotips.matchingNotifications.matchesTable.onFailureAlert')) " + failedIds.join());
```
allasm: +1, phenotips.matchingNotifications.matchesTable.onFailureAlert is present in ApplicationResources.properties

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/unnotifiedMatchesTable.js
```diff
@@ -146,18 +146,12 @@ var PhenoTips = (function (PhenoTips) {
...
         var [successfulIds, failedIds] = this._utils.getResults(results);
 
         if (failedIds.length > 0) {
-            var operation = reject ? "Setting status" : "Status setted";
-            if (failedIds.length == 1) {
-                alert(operation + " match failed.");
-            } else {
-                alert(operation + " matches with the following ids failed: " + failedIds.join());
-            }
+            alert("Setting " + status + " status for matches with the following ids failed: " + failedIds.join());
```
allasm: +1, though we should do error reporting not through `alert()`-s

matching-notification-api/src/main/resources/ApplicationResources.properties
```diff
@@ -2,9 +2,7 @@ phenotips.matching.ajaxutils.requestSent=Sending request...
 phenotips.matching.ajaxutils.done=Done
```
allasm: I don't see how this commit is related to VS, all changes seem to be relevant for matching notification and exomiser (both already present without VS), not VS. Commit title should be adjusted to reflect that

similarity-data-impl/src/main/resources/ApplicationResources.properties
```diff
@@ -1,16 +1,73 @@
...
-
 phenotips.similarCases.modeOfInheritance=Mode of inheritance
-
 phenotips.similarCases.phenotypicFeaturesBreakdown=Phenotypic features breakdown
-
 phenotips.similarCases.geneMatchingBreakdown=Gene matching breakdown
-
 phenotips.similarCases.showDisclaimer=Show disclaimer
 phenotips.similarCases.disclaimerDialogTitle=Service Disclaimer
+phenotips.similarCases._UNKNOWN_INFO_MARKER=unknown
```
allasm: Why name these 4 strings differently form all others, in capitals with an underscore?

ui/src/main/resources/data/GrantMatchAccess.xml
```diff
@@ -42,10 +42,9 @@
   #set ($accessRequest = $services.anonymousCommunication.grantAccess($request.id))
```
allasm: Just curious if we ever use this file (`GrantMatchAccess.xml`), I couldn't find where it is used?

veronikaslc: not sure about this file as well

similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
```diff
@@ -73,9 +73,6 @@
     /** The number of records to fully score. */
     private static final int SEED_QUERY_SIZE = 50;
```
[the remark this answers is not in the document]
sdumitriu: Hm, small number of seeds was introduced to help performance when there was only the in-patient-record instantly-computed similar cases. This needed a fast computation time, since users don't want to wait 5 minutes to see similar patients. But the way we compute and display matches has changed since then, so we need to re-evaluate this.

For all matches table, we don't care if the computation takes longer, since it's something we already expect to take a lot of time and we're fine with it taking a whole night.

@allasm For the in-patient-record similar cases, do we still compute matches each time the page is loaded?

We could improve the pool of seeds in three ways:
1. Increase the number of seeds
2. Limit the results by a Solr score threshold instead of by absolute count
3. Don't seed at all, but instead compute everybody-vs-everybody scores, which will make it impossible to miss similar cases (we have an open issue about missed similar cases because of slightly different phenotypes, and will also make exome-only matching possible); this is the costliest, but if it doesn't slow down things too much, now that we have incremental searches, may be the best option

similarity-data-impl/src/main/resources/ApplicationResources.properties
veronikaslc: no particular reason, I guess

similarity-data-rest/src/main/java/org/phenotips/data/similarity/rest/internal/DefaultPatientMatchResource.java
```diff
@@ -0,0 +1,165 @@
...
+            return Response.status(Response.Status.BAD_REQUEST).build();
+        }
+        final Patient patient = this.repository.get(reference);
+
+        if (patient == null) {
+            this.logger.error("Patient with ID: {} could not be found.", reference);
+            return Response.status(Response.Status.BAD_REQUEST).build();
+        }
+        final List<PatientSimilarityView> matches = this.similarPatientsFinder.findSimilarPatients(patient);
+        final boolean isManager = ((SecurePatientAccess) this.manager.getPatientAccess(patient)).hasAccessLevel(MANAGE);
```
allasm: Matches are not saved to the matching notification table here, but they should be. A good fix is to use `LocalMatcher` or even higher-level match-finding code to find matches here, so that all the low-level work implemented there (such as saving matches) is reused

matching-notification-api/src/main/resources/ApplicationResources.properties
```diff
@@ -3,6 +3,7 @@ phenotips.matching.ajaxutils.done=Done
 phenotips.matching.ajaxutils.failure=Request failed
 
 phenotips.matchingNotifications.invalidScore=Score must be a number between 0 and 1
+phenotips.matchingNotifications.success=Success
```
allasm: not used anywhere

matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/PatientInMatch.java
```diff
@@ -141,4 +141,10 @@
      * @return What type of access does the user have to this patient profile.
      */
     AccessLevel getAccess();
+
+    /**
+     * @return status of genes (solved or candidate) stored in PatientGenotype class,
+     *         or null if patient does not have genotype.
```
allasm: I think we should add "...or genotype status is unknown", since this is what we do for genes in MME patients

matching-notification-api/src/main/resources/ApplicationResources.properties
```diff
@@ -41,31 +41,31 @@ phenotips.matchingNotifications.table.accessFilter=Access level:
 phenotips.matchingNotifications.table.owner=patients owned by me
 phenotips.matchingNotifications.table.view=patients shared with me
 phenotips.matchingNotifications.table.notifiedStatusFilter=Notified status:
-phenotips.matchingNotifications.table.notified=notified
+phenotips.matchingNotifications.table.notified=Search for notified only
 phenotips.matchingNotifications.table.unnotified=unnotified
```
allasm: this one is no longer used

allasm:
Current problem: PatientMatchEmailNotifier.java

createEmails(List<PatientMatch> matches, Map<Long, List<String>> matchesIds) is a bad API which can't be used to notify non-local patients. It has to be rewritten completely. We did not run into this before because we just happened to use self-matches which hepen to be valid local IDs even when they were actually remote IDs

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/MatchingNotificationResource.java
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * Resource for working with patients matches and notifications.
+ *
+ * @version $Id$
+ * @since 1.1
+ */
+@Unstable("New API introduced in 1.1")
+@Path("/patients/matching-notification")
+@Relation("https://phenotips.org/rel/matchingNotification")
+@ParentResource(PatientResource.class)
```
sdumitriu: Should be `PatientsResource`, plural.

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/MatchingNotificationResource.java
```diff
@@ -0,0 +1,153 @@
...
+     * <dl>
+     * <dt>ids</dt>
+     * <dd>a JSON object (in the format as described above) with a list of match ids that given patients ids
+     * should be notified about</dd>
+     * </dl>
+     *
+     * @return result JSON
+     */
+    @POST
+    @Produces(MediaType.APPLICATION_JSON)
```
sdumitriu: Needs `@Consumes`?

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/MatchingNotificationResource.java
```diff
@@ -0,0 +1,153 @@
...
+     * Input: ["1", "2"], "saved"
+     * Output: {"results": {"success": [1,2]}}
+     * </pre>
+     *
+     * @param matchesIds List of matches IDs to change the status for
+     * @param status whether matches status should be set as 'saved', 'rejected' or 'uncategorized'
+     * @return result JSON
+     */
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
```
sdumitriu: Does it consume anything? If form params are expected, then the input should be `form-encoded`, not `JSON`.

matching-notification-ui/src/main/resources/MatchingNotification/MatchesTable.xml
```diff
@@ -0,0 +1,568 @@
...
+    &lt;/div&gt;##--closing table div id="table-container"
+  &lt;/div&gt;
+&lt;/div&gt;
+{{/html}}
+{{/velocity}}</content>
+<object>
+    <name>Families.WebHome</name>
+    <number>0</number>
+    <className>XWiki.UIExtensionClass</className>
+    <guid>e0686a4d-61ec-4d99-9941-fb0041842158</guid>
```
sdumitriu: Don't just copy GUIDs, they should be unique. Same for the other objects.

matching-notification-ui/src/main/resources/MatchingNotification/MatchesTable.xml
```diff
@@ -0,0 +1,568 @@
...
+      &lt;/span&gt;
+      &lt;div id="send-notifications-messages"&gt;&lt;/div&gt;
+    &lt;/div&gt;
+    &lt;/div&gt;##--closing table div id="table-container"
+  &lt;/div&gt;
+&lt;/div&gt;
+{{/html}}
+{{/velocity}}</content>
+<object>
+    <name>Families.WebHome</name>
```
sdumitriu: Wrong name.

similarity-data-rest/src/main/java/org/phenotips/data/similarity/rest/PatientMatchResource.java
```diff
@@ -0,0 +1,63 @@
...
+import javax.ws.rs.core.Response;
+
+/**
+ * Resource for working with patient match data.
+ *
+ * @version $Id$
+ * @since 1.1
+ */
+@Unstable("New API introduced in 1.1")
+@Path("/patients/{patient-id}/similar-cases")
```
sdumitriu: For 1.2 only, this will have to be updated to `{entity-id}`.

ui/src/main/resources/data/GrantMatchAccess.xml
sdumitriu: It was used; in the email sent from the patient record's similar cases, the message was something like "hello, we have similar patients, here's my patient info, if you would like to collaborate, click here to grant me access to your patient". I see that is no longer happening, so this is probably unused.

matching-notification-ui/src/main/resources/MatchingNotification/MatchesTable.xml
allasm: we will remove this object and replace with a WebHome for this space

similarity-data-rest/src/main/java/org/phenotips/data/similarity/rest/PatientMatchResource.java
allasm: does it affect anything but naming? Any reason not to do it right away?

sdumitriu: This code itself will work, but what won't work is linking to/from the patient JSON. The autolinker creates links from the patient record to this "child" resource, and from this resource to the parent. Since the parent `/patient/` resource has a parameter, computing the URL requires knowing what is the value for that parameter, and since the parameters have different names, the autolinker won't know that the value for `entity-id` is supposed to be used for `patient-id`. The reverse is also true when linking from the patient to this resource.

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -923,6 +923,10 @@ var PhenoTips = (function (PhenoTips) {
                 this._contactContainer.down('textarea[name="message"]').update('');
                 this._contactContainer.down('input[name="subject"]').value = '';
                 this._contactContainer.down('input[name="to"]').value = '';
+                // clone send button to remove all event handlers
```
allasm: if this works lets keep it for now, but this looks hack-ish, what event handlers do we try to remove and why?

PR #188, opened 2018-03-19
PN-292: My Matches page shows disabled remote servers as available filters

sdumitriu:
Fixed

### DIFF
--- a/matching-notification-ui/src/main/resources/MatchingNotification/MatchesTable.xml
+++ b/matching-notification-ui/src/main/resources/MatchingNotification/MatchesTable.xml
@@ -111,7 +111,7 @@ $xwiki.jsx.use('PhenoTips.Widgets')##
                 #set ($preferenceDoc = $xwiki.getDocument("XWiki.XWikiPreferences"))
                 #set ($allRemotes = $preferenceDoc.getObjects("PhenomeCentral.RemoteMatchingServiceConfiguration"))
                 #foreach ($config in $allRemotes)
-                  #if ($config.getProperty('serverId').getValue() != "" &amp;&amp; $config.getProperty('humanReadableName').getValue() != "")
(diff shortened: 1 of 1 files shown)


allasm:
In this filter users only see servers which are involved in at least one of their matches. Should we do the same for admins, then this PR wont be necessary? In fact, this will be better, as if a server is disabled after some matches are found, with this fix it still will not be listed. So while PR itself is correct, I'm not sure this is what we want. Lets discuss at devmeeting

veronikaslc:
Here are my 5c: 
Table shows disabled remote servers as available filters because the matches are there and behavior with logic is correct. If we show data - we show filter for it.
So it seems to me the problem is - we show matches from disabled servers. If this is indeed the problem - the fix is different.

sdumitriu:
After loading the matches, are filters applied?

If yes, then since there are no checkboxes for disabled servers, then the matches from those servers will be filtered out -> no bug.

If no, then there will be a jump in results count when the first filter is applied -> bug.

Can @veronikaslc confirm that filters are applied right after the rows are received before displaying the results?

veronikaslc:
@sdumitriu  This is correct, filters are applied after matches are loaded before building a table. 
They work as follows: first filer checkboxes are generated with value=serverID and checked properties if we want to show matches from those servers and unchechecked if not, hidden if we do not want to show filter itself. Then on page load checkboxes are parsed and saved into filter json object `filter` where server ID corresponds to the boolean value, true if box checked, false if not. Then on matches load or on filter change the filter is apllied, whether to show match based on reading boolean value from filter json, like ` filter[match.serverID]`.
Given that , not generating filter checkboxes at all may potentially break javascript. I'd generate hidden unchecked checkboxes for disabled servers. This is safier and those matches will be always filtered out.

PR #189, opened 2018-03-19
Changes for 1.1.10

veronikaslc:


matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -62,26 +63,15 @@ var PhenoTips = (function (PhenoTips) {
...
 
-        this._contactContainer = this._generateContactDialogContainer();
-        this._contactDialog = new PhenoTips.widgets.ModalPopup(this._contactContainer, false, {'title': this._CONTACT_DIALOG_TITLE, 'verticalPosition': 'top'});
+        this._contactDialog = new PhenoTips.widgets.MatcherContactDialog();
 
         // memorise filers open/close state
         var filtersButton = $('toggle-filters-button');
-        var key = this._utils._getCookieKey(filtersButton.up('.entity-directory').id);
+        var key = this._utils.getCookieKey(filtersButton.up('.entity-directory').id);
         this._toggleFilters(filtersButton.up('.xwiki-livetable-topfilters-tip'), (XWiki.cookies.read(key) == 'hidden'));
```
allasm: unrelated to your changes, but if we don't use "livetable" any more (we do not, right?) maybe we should not care about livetable cookies as well?

veronikaslc: Cookies hold matching table filters expand/collapse state. We use livetables CCS classes for filters and pagination.

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/MatchingNotificationResource.java
```diff
@@ -175,4 +175,22 @@ Response getEmailToBeSent(@FormParam("matchId") String matchId,
...
+     * @param referenceServerId id of the server that hosts reference patient
+     * @param matchedPatientId id of matched patient to load matches for
+     * @param matchedServerId id of the server that hosts matched patient
+     * @return a single matchID
+     */
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/get-matchid")
+    Response getMatches(@FormParam("referencePatientId") String referencePatientId,
```
allasm: the method is named `getMatches()`, but URL is `get-matchid`, and it actually returns only match ID, not a list of matches (so the comment is incorrect)

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/internal/DefaultMatchingNotificationResource.java
```diff
@@ -455,4 +456,38 @@ private boolean isNonexistingOrSelfMatch(PatientMatch match)
...
+            return Response.status(Response.Status.BAD_REQUEST).build();
+        }
+
+        List<PatientMatch> matches = this.matchStorageManager.loadMatchesBetweenPatients(
+            referencePatientId, referenceServerId, matchedPatientId, matchedServerId);
+
+        JSONObject matchIDJson = new JSONObject();
+
+        if (!matches.isEmpty()) {
+            matchIDJson.put(MATCHID_STRING, getLastOutgoingMatchId(matches));
```
allasm: I don't remember by now, but what if there are matches, but none of them are outgoing, should we return `null` or the id of the last incoming match instead? I forgot, where do we use this endpoint?

P.S. also I don't think we should be returning `null`-s, but thats what we do in `getLastOutgoingMatchId()` if there are no outgoing matches

allasm:
@sashaandjic re: PN-317: An internal error when trying to send admin email notifications

There was a bug in the code which resulted in null pointer exception and which prevented all selected email from being sent. I fixed the null pointer and added a log message which should help us diagnose the error in the future, but an error may still occur (though it will affect only one email now). The problem is while I know where that null pointer happens, it seems the conditions for it to happen are impossible and I was unable to reproduce the bug locally. So it is important to test this well both locally and on preprod before we deploy it to production

matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
```diff
@@ -260,6 +274,72 @@ public boolean saveRemoteMatches(List<? extends PatientSimilarityView> similarit
...
+                + "or (accessObj.className = 'PhenoTips.CollaboratorClass' and accessProp.value in "
+                + "(" + userEntityListForSQL + ")) "
+                + "or (accessObj.className = 'PhenoTips.VisibilityClass' and accessProp.value in ('public', 'open')))",
+                org.xwiki.query.Query.XWQL);
+            List<String> rawDocNames = q.execute();
+
+            Set<String> patientNames = new HashSet<>();
+            for (String wikiDocName : rawDocNames) {
+                // remove the space prefix, as it is not recorded as part of a match
+                patientNames.add(wikiDocName.replaceFirst("^[^.]+\\.", ""));
```
sdumitriu: This is unnecessary. Instead of selecting `doc.fullName`, select `doc.name`, which is already what you want.

matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
```diff
@@ -260,6 +274,72 @@ public boolean saveRemoteMatches(List<? extends PatientSimilarityView> similarit
...
+            }
+
+            // FIXME: to be removed after testing is complete. In some cases it is easier to spot errors in this
```
sdumitriu: Is testing complete now? You can remove this log line.

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -472,8 +413,14 @@ var PhenoTips = (function (PhenoTips) {
         return summary;
     },
 
+// FORMATTING MATHCHES BEFORE TABLE BUILD
```
sdumitriu: Typo: `matHches`.

similarity-data-impl/src/main/resources/ApplicationResources.properties
```diff
@@ -1,14 +1,14 @@
 phenotips.similarCases.title=Similar cases
-phenotips.similarCases.subtitle=Matches with patient records in this database
+phenotips.similarCases.subtitle=Matches in phenomecentral
```
sdumitriu: Please use `PhenomeCentral`, not `phenomecentral`.

matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
allasm: Aha

matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
allasm: Not yet. Need to see what happens on `preprod` first

PR #194, opened 2018-08-23
Matching table improvements (PN-326, PN-336, PN-342) and compatibility fixes with PhenoTips 1.4 (PN345, dependencies)

veronikaslc:
Added sorting by match found timestamp in descending order in UI


veronikaslc:
For easier testing added PN-336 fix here.

matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
```diff
@@ -513,15 +528,22 @@ private JSONObject getOwnership()
...
                 Set<Group> userGroups = GROUP_MANAGER.getGroupsForUser(currentUser);
+                XWikiContext context = CONTEXT_PROVIDER.get();
                 for (Group group : userGroups) {
                     DocumentReference groupRef = group.getReference();
                     if (groupRef.equals(ownerRef)) {
                         userGroupIsOwner = true;
+                        break;
+                    }
+
+                    if (userIsGroupManager(userRef, groupRef, context)) {
```
allasm: this does not find if current user is the administrator of the match owner's group, it just finds that current user is administrator of some group

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.css
```diff
@@ -245,3 +245,10 @@ input.mme {
   max-width: 180px; /* based on the width of filters column, defined to be 240px in DBWebHomeSheet */
   vertical-align: top;
 }
+.fa.fa-sort, .fa.fa-sort-up, .fa.fa-sort-down {
```
allasm: it is better to have a custom class, instead of overwriting generic `.fa`

matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
allasm: ...btw, I think to speed things up for admins (who get to see 1000s of matches) we can say that if current user is admin `userIsGroupManager` is always true regardless of actual match owner. That way for admins we will never run the expensive "is group admin" check, and for users who only see a few matches it is not  anbig deal to run it

veronikaslc:
@allasm codereview comments addressed

veronikaslc:
@sashaandjic @marta- A question regarding PN-335: if a group administrator is contacting about a match, the email will say: "PhenomeCentral has identified similarities between one of my patients and your patient." But the actual owner of the patient will be a different user. Do we need to change wording in this case? Do we need to CC the actual owner?

sashaandjic:
If we replace "my" with "our" would that cover both cases?
I would cc the owner of the patient record.

sashaandjic:
@veronikaslc yesterday while testing this PR I was not able to see any local matches when I was logged as an administrator. I my database I have 3 user accts and plenty of local matches for all 3.

veronikaslc:
@sashaandjic to see all available matches you have to select filter "group" checkbox at the very bottom if filtering panel, its unselected by default which confuses. @allasm may be we should make it selected by default?
In any case lets see into it alltogether.

sashaandjic:
it doesn't work with group unchecked
screen shot 2018-09-20 at 1 03 18 pm

here is the screen shot of the same page when logged as a user
screen shot 2018-09-20 at 1 03 51 pm

veronikaslc:
@sashaandjic If you a logged in as a goup administrator, you can see a match only in the following 3 cases:
- if you are a collaborator to one of the patients in the match
- if one of the patients in the match has open visibility
- if one of the patients in the match has public visibility
In all other cases you will see no matches.

veronikaslc:
@marta- @sashaandjic It does not seem that we restrict sending notification email from the patient form anyhow other than checking that the user has view rights to the patient [here](https://github.com/phenotips/patient-network/blob/stable-1.1.x/matching-notification-api/src/main/java/org/phenotips/matchingnotification/internal/DefaultMatchingNotificationManager.java#L115)
Other comments discussed today (adjusting email template and CC owner) are addressed in last 3 commits and ready to be tested.

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -99,9 +99,42 @@ var PhenoTips = (function (PhenoTips) {
             this._showMatches();
         }
 
+        // defines sorting order for UI column sorting elements events listeners,
+        // values: 1 if descending, -1 if ascending
+        this._scoreSortingOrder = 1;
```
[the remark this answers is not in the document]
allasm: OK, i missed that changing line, and I guess it is ok to leave as is, but the only use of specifically `1` and `-1` is when you multiply the value by `-1` to invert. Could have as well just explicitly wrote (which is also more clear) something like
`value = (value == "ascending") ? "descending" : "ascending"` which would make it immediately clear the sorting order is reversed, as opposed to trying to see why we do some multiplications there

allasm:
Hmm, I think cases where you are a collaborator can be treated as "other" (so that checkbox should be shown for non-admin users as well). Alternatively we can add a "collaborator" checkbox, or treat collaboration as "my own"

sashaandjic:
I would use "My group" filter for the collaborative assignments explicit and implicit (via group). Maybe "My group" should be changed into "Shared with me" or "My consortium"?

allasm:
actually, after thinking a bit, patients where the user is a collaborator is by definition of the issue "other matches". The initial issue was about seeing too many matches, including matches where the current user is not an owner of either match. For non-admin users the only matches they see where they are not owner is where they are collaborators or their group is an owner. Group has its own checkbox, so matches where the user is collaborator are "other", since there is nothing else a regular user might see

similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
```diff
@@ -218,8 +223,7 @@ private boolean filterPatient(Patient matchPatient, Family family, EntityReferen
             return true;
         }
         // filter out patients with access level less than defined access level threshold
-        AccessLevel patientAccessLevel =
-            this.accessHelper.getAccessLevel(matchPatient, this.accessHelper.getCurrentUser());
+        AccessLevel patientAccessLevel = this.permissions.getPatientAccess(matchPatient).getAccessLevel();
```
allasm: I approve the change but object the commit message.

Commit message says "I switched from using class OLD to using class NEW, because NEW checks X and OLD does not check X but checks Y instead".

From that a code reviewer can derive that before we only checked for Y but now we check for X instead. In fact, NEW checks for X and then internally calls OLD to check for Y, so now we check for both X and Y, which is totally not clear form the commit message without digging deep into the code

veronikaslc: The commit message does not have the word `instead`. It clearly says that methods `return different results`. `DefaultPatientAccess` considers owner or Admin cases, while `PatientAccessHelper` does not consider those situations but runs checks of access levels for all groups user belongs to. No other words about what `DefaultPatientAccess` does or does not.

allasm: you have the words `but runs checks` which normally means "instead". E.g. "I don't know how to play piano, but i know how to play violin". While in the example above, both of them run those checks

veronikaslc: `but` was entirely in the part about what PatientAccessHelper does, not DefaultPatientAccess. Thats why `but` was not for comparing two methods.
There the sentence was constructed as: "Method X does not do THIS but does THAT." 
PatientAccessHelper `does not consider those situations` but `runs checks of access levels for all groups user belongs to transitionally`. And it literally does not consider owner and Admin, but what it does is runs other checks for groups.

allasm: Your statement is correct. But put together like that it does imply that one does one thing while the other does the other (while, again, in fact one does both while the other does just one)

IMHO a better way to phrase that would simple be `switched form using OLD to NEW because OLD does not check case X` which explains the problem and the fix in the most concise way. Or with more explanations: `switched form using OLD to NEW because NEW internally calls OLD but also checks X which OLD does not check` or `switched form using OLD to NEW because OLD does not check case X while NEW checks X and otherwise works like OLD` which makes it crystal clear that most cases will stay the same, while cases which match X are handled differently (and only those cases)

veronikaslc: Well, the commit message says `Used PermissionsManager instead of PatientAccessHelper to compute AccessLevel, because they return different results (...)` and in `()` I gave exact difference on the two methods: considering or not owner and admin cases. Statements are correct. The understanding and further investigation of PT code is up to reader and not relevant to this particular fix.

veronikaslc: It doesn't imply, because `but` was not between methods comparing them. It was only inside the part about what PatientAccessHelper does.

allasm: not exactly, the phrase was about both of them: `X does A, Y does not do A but Y does B`. It sure implies "X does A, but not B while Y does B but not A".

If you said "X does A and calls Y, which does not do A but does B" it would be clear who do and doesn't do what

veronikaslc: Also, I was no explaining how X works. I said "X considers A while Y does not consider A" rest is lyrics :)

allasm: Our context is not abstract logical statements. Our context is making sense of some changes, where Y was replaced by X. So code-reviewers naturally need to know not how Y works in a vacuum, but how is Y different from X. In this context saying something about Y in the style "...but does that" kind of implies this is that difference

veronikaslc: IMHO there is nothing to discuss or object. The commit message contains the exact pointed reason of a change - the key difference in consideration of owner and admin cases explained in the most clear way- and provides more than enough literally correct information for reviewer. It does not have to explain how PT code works in every detail and it does not have to explain every other difference or similarity in PT code methods. Once again - this change was done because of that exact difference in methods, when one considers owner and admin case, and other does not. All other topics together with all assumptions about how exactly methods work are simply out of the scope of my commit and out of its context.

veronikaslc: I agree, I should not have said `but` ending.

PR #196, opened 2018-09-13
PN-335: Group adminitrators should be able to contact case owners of any matched record from their group

veronikaslc:


veronikaslc:
Closed in favour https://github.com/phenotips/patient-network/pull/194

PR #197, opened 2018-09-20
PN-339: Create an API method to get the total number of matches

veronikaslc:


matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
```diff
@@ -565,4 +574,20 @@ private void saveMatches(Session session, List<PatientMatch> matches)
             session.save(match);
         }
     }
+
+    @Override
+    public Long getNumberOfRemoteMatches()
+    {
+        Session session = this.beginTransaction();
+        long result = -1;
```
allasm: it is better to use `null` than -1, after all `Long` is an object which supports this special value

matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
```diff
@@ -110,10 +110,14 @@
...
      * */
     private static final String HQL_QUERY_FIND_ALL_MATCHES_BY_SCORE =
-            "from DefaultPatientMatch as m where"
+        "from DefaultPatientMatch as m where"
             + " score >= :minScore and phenotypeScore >= :phenScore and genotypeScore >= :genScore and"
             + " (m.notified = true or (not exists (" + HQL_QUERY_SAME_PATIENT_BUT_NOTIFIED + ")))";
 
+    /** A query used to get the number of all remote matches. */
+    private static final String HQL_QUERY_REMOTE_MATCHES =
+        "select count(*) from DefaultPatientMatch as m where m.referenceServerId = '' or m.matchedServerId =''";
```
allasm: this will return a local match which has both of those servers as blank. I think what you mean is `!= ''`

PR #198, opened 2018-09-24
PN-337: Optimize patients filtering for match finding.

veronikaslc:


matching-notification-api/src/main/java/org/phenotips/matchingnotification/finder/internal/DefaultMatchFinderManager.java
```diff
@@ -60,30 +60,43 @@
...
+                    + "visibility.visibility <> 'private' and "
+                    // filter by last update date
+                    + "doc.contentUpdateDate > :startTime and "
+                    + "order by patient.identifier desc",
                 Query.XWQL);
+            if (onlyCheckPatientsUpdatedAfterLastRun) {
+                q.bindValue("startTime", System.currentTimeMillis());
+            } else {
+                q.bindValue("startTime", 0);
+            }
```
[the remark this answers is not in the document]
veronikaslc: @marta- @sdumitriu wdyt?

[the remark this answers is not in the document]
allasm: also, e.g. visibility - we may want to check "user access" in a more complicated way than just checking "visibility" later, e.g. consider user access to a project or study a patient belongs to - well, or whatever else comes up in the future. A java "manager" will take care of it automatically, a query will not. And I'm one of the first to point out PT/PC runs slowly, but this part did not seem to be the bottleneck, so is not a good candidate for moving Java code into a query

matching-notification-api/src/main/java/org/phenotips/matchingnotification/finder/internal/DefaultMatchFinderManager.java
```diff
@@ -60,30 +60,41 @@
...
+                    + "doc.object(PhenoTips.VisibilityClass) as visibility "
+                    + "where patient.identifier is not null and "
+                    // exclude solved patients
+                    + "patient.solved <> '1' and "
+                    // filter by last update date
+                    + "doc.contentUpdateDate > :startTime and "
+                    + "order by patient.identifier desc",
                 Query.XWQL);
+            if (onlyCheckPatientsUpdatedAfterLastRun) {
+                q.bindValue("startTime", System.currentTimeMillis());
```
allasm: There is a problem here. `onlyCheckPatientsUpdatedAfterLastRun` means not patient updated after `System.currentTimeMillis()`, but updated after that particular matcher was run, which only that matcher knows (see line https://github.com/phenotips/patient-network/pull/198/commits/a31b758a238635fbc477b0894d0210d0506a4f71#diff-9a72985fda0ed745264848cb16cafa2fR123)

So the "is patient updated since last time" check should be done within the matcher and not here, sinc eit is matcher dependant

similarity-search/src/checkstyle/checkstyle-suppressions.xml
```diff
@@ -0,0 +1,28 @@
...
+-->
+
+<!DOCTYPE suppressions PUBLIC
+    "-//Puppy Crawl//DTD Suppressions 1.0//EN"
+    "http://www.puppycrawl.com/dtds/suppressions_1_0.dtd">
+
+<suppressions>
+  <suppress checks="ClassFanOutComplexity" files="SolrSimilarPatientsFinder"/>
+  <suppress checks="ReturnCount" files="SolrSimilarPatientsFinder"/>
+</suppressions>
```
allasm: I think we should check if inline style suppression ( `@SupressXXX`) works before adding a file here

similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
```diff
@@ -177,6 +194,39 @@ public int compare(PatientSimilarityView o1, PatientSimilarityView o2)
...
 
+    private boolean filterPatient(Patient matchPatient, Family family, EntityReference refOwner,
+        String requiredConsentId)
+    {
+        if (matchPatient == null) {
+            // Leftover patient in the index, should be removed
+            return true;
+        }
+        // filter out patients from the same family
+        if (family != null && family.isMember(matchPatient)) {
```
allasm: This does not work correctly for remote patients. I just checked how `isMember()` works (see https://github.com/phenotips/phenotips/blob/master/components/family-studies/api/src/main/java/org/phenotips/studies/family/internal/PhenotipsFamily.java#L140) - it just compares patient IDs. Remote patients may have an ID that matches an ID of some family member, but it still does not mean that patient is in a local family. Need to add a check first that the patient is local

veronikaslc: @allasm `matchedPatient` can't possibly be remote. For reference remote patients family is null.

matching-notification-api/src/main/java/org/phenotips/matchingnotification/finder/internal/DefaultMatchFinderManager.java
veronikaslc: Since last update info is stored in a separate document in phenomecentral space it does not depend on a matcher. Also we do not have any manager public method to use to get last update time, so it can be done here in the match finder manager. I will update PR.

similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
allasm: hmm, right, my bad. I will remove this comment then

similarity-search/src/checkstyle/checkstyle-suppressions.xml
veronikaslc: It doesn't.

allasm:
It will be interesting to compare performance of local match finder on `preprod` before and after the change

similarity-search/src/checkstyle/checkstyle-suppressions.xml
sdumitriu: It only works for PT 1.4, but this is the stable branch still using 1.3.

similarity-search/src/main/java/org/phenotips/similarity/internal/SolrSimilarPatientsFinder.java
```diff
@@ -177,6 +194,39 @@ public int compare(PatientSimilarityView o1, PatientSimilarityView o2)
...
+            return true;
+        }
+        // filter out patients of the same owner
+        if (refOwner.equals(this.accessHelper.getOwner(matchPatient).getUser())) {
```
sdumitriu: This may trigger a `NullPointerException` if the owner is `null`, which shouldn't happen but as we've seen yesterday, it may if there are other bugs.

PR #199, opened 2018-09-28
PN-340: Users with edit rights should be able to send notifications from Notification matches page

veronikaslc:


matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -844,7 +844,10 @@ var PhenoTips = (function (PhenoTips) {
     _getNotified : function(record)
     {
         var td = '<td style="text-align: center" name="contacted">';
-        if (!this._isAdmin && (record.matched.isOwner || record.reference.isOwner) && (this._utils.isBlank(record.matched.pubmedId) || this._utils.isBlank(record.reference.pubmedId))) {
+        var accessAboveEdit = record.matched.access == "edit"   || record.reference.access == "edit"
+                           || record.matched.access == "owner"  || record.reference.access == "owner"
+                           || record.matched.access == "manage" || record.reference.access == "manage";
+        if (!this._isAdmin && accessAboveEdit && (this._utils.isBlank(record.matched.pubmedId) || this._utils.isBlank(record.reference.pubmedId))) {
```
allasm: should we be a bit more specific here about the solved cases? This code still lets someone to notify the owner of  a solved record of a match. Do we want that @marta- ? Should it be if `this._utils.isBlank(record.matched.pubmedId) **&&**  this._utils.isBlank(record.reference.pubmedId)` instead of **||**?

PR #205, opened 2018-11-01
RM-122: Parse remote patient ID for MyGene2

veronikaslc:
Parse remote patient ID because MyGene2 uses URLs in the ID field to ensure that the public MyGene2 profiles are actually delivered to the end user

### DIFF
--- a/matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
+++ b/matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
@@ -793,7 +793,14 @@ var PhenoTips = (function (PhenoTips) {
             td += '<a href="' + patientHref + '" target="_blank" class="patient-href">' + patient.patientId + externalId + '</a>';
         } else { // remote patient
             // TODO pass a server name in JSON as well to display a server name instead if server ID in the table
-            td += '<label class="patient-href">' + patient.patientId + externalId + ' (' + patient.serverId + ')</label>';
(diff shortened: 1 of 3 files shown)


veronikaslc:
Rebased PR to use generic approach

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -793,7 +793,14 @@ var PhenoTips = (function (PhenoTips) {
             td += '<a href="' + patientHref + '" target="_blank" class="patient-href">' + patient.patientId + externalId + '</a>';
         } else { // remote patient
             // TODO pass a server name in JSON as well to display a server name instead if server ID in the table
-            td += '<label class="patient-href">' + patient.patientId + externalId + ' (' + patient.serverId + ')</label>';
+            // Parse remote patient ID because MyGene2 uses URLs in the ID field to ensure that the public MyGene2 profiles are actually delivered to the end user
+            if (patient.serverId == "mygene2" && patient.patientId && patient.patientId.startsWith("http")) {
+                var matches = patient.patientId.match(/[\d]+/g);
```
allasm: I would add a comment to explicitly describe/confirm the logic as it is not obvious that what is happening is actually correct ("find the last group of digits in the URL and use it as an ID")

allasm:
PR seems to be correct, but we need to wait for an answer from MyGene2 to know if our logic is future poof

("1. How likely is it for the patient ids to be anything other than numbers?
2. How likely is it for the URL structure to change?)

PR #212, opened 2018-11-09
PN-355: Matching notifications page: By default have rejected option unchecked; PN-354: Matching notification page: modify default settings for "Matches for patients owned by"

veronikaslc:
Done.

### DIFF
--- a/matching-notification-ui/src/main/resources/MatchingNotification/MatchesTable.xml
+++ b/matching-notification-ui/src/main/resources/MatchingNotification/MatchesTable.xml
@@ -144,8 +144,8 @@ $xwiki.jsx.use('PhenoTips.Widgets')##
               &lt;/li&gt;
               &lt;li&gt;$escapetool.xml($services.localization.render('phenotips.matchingNotifications.table.ownershipFilter'))&lt;br&gt;
                 &lt;label&gt;&lt;input name="ownership-filter" type="checkbox" value="me" checked&gt; $escapetool.xml($services.localization.render('phenotips.matchingNotifications.table.owned_by_me'))&lt;/label&gt;&lt;br&gt;
-                &lt;label&gt;&lt;input name="ownership-filter" type="checkbox" value="group" checked&gt; $escapetool.xml($services.localization.render('phenotips.matchingNotifications.table.owned_by_group'))&lt;/label&gt;&lt;br&gt;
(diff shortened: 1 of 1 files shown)


matching-notification-ui/src/main/resources/MatchingNotification/MatchesTable.xml
```diff
@@ -138,14 +138,14 @@ $xwiki.jsx.use('PhenoTips.Widgets')##
...
               &lt;/li&gt;
               &lt;li&gt;$escapetool.xml($services.localization.render('phenotips.matchingNotifications.table.ownershipFilter'))&lt;br&gt;
                 &lt;label&gt;&lt;input name="ownership-filter" type="checkbox" value="me" checked&gt; $escapetool.xml($services.localization.render('phenotips.matchingNotifications.table.owned_by_me'))&lt;/label&gt;&lt;br&gt;
                 &lt;label&gt;&lt;input name="ownership-filter" type="checkbox" value="group" checked&gt; $escapetool.xml($services.localization.render('phenotips.matchingNotifications.table.owned_by_group'))&lt;/label&gt;&lt;br&gt;
-                &lt;label&gt;&lt;input name="ownership-filter" type="checkbox" value="others" #if(!$hasAdmin)checked#end&gt; $escapetool.xml($services.localization.render('phenotips.matchingNotifications.table.owned_by_other'))&lt;/label&gt;&lt;br&gt;
+                &lt;label&gt;&lt;input name="ownership-filter" type="checkbox" value="others" checked&gt; $escapetool.xml($services.localization.render('phenotips.matchingNotifications.table.owned_by_other'))&lt;/label&gt;&lt;br&gt;
```
sdumitriu: Isn't the intended behavior to have this checked **only** for admins? That means simply removing the `!` from the condition, not the whole condition.

sashaandjic:
I would keep Other option turned on by default for everyone. When Brittney logged this task, her main concern were numerous matches where on one side had a public patient. With Public option turned off by default, they will go away. I believe having ability to see Other by default is beneficial as you will see matches for all the patient cases you have access to. Users can always turn off this option if they want to see only cases owned by them.

sashaandjic:
After talking to Brittney, we realized Other should be turned off by default for typical users.
Global admins and group admins should have it turned on by default.

veronikaslc:
Addressed codereview comments.

veronikaslc:
Re-tested & re-approved by Sasha.

PR #216, opened 2018-11-14
PN-358: Add an info button next to the filtering options under "Matches for patients owned by:"

veronikaslc:


sashaandjic:
Brittney's suggestions:

allasm:
Has this been discussed? I recall in the past we discussed that users may be confused that some of their matches are not shown (and may not notice this checkbox)

sashaandjic:
Yes, we discussed showing uncontacted matches by default.

veronikaslc:
Removed PN-359 from this PR into separate branch `PN-359`

PR #217, opened 2018-11-15
PN-323: On My Matches page make match score search criteria more user friendly

veronikaslc:
Done.

### DIFF
--- a/matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
+++ b/matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
@@ -41,13 +41,9 @@ var PhenoTips = (function (PhenoTips) {
         $('send-notifications-button').on('click', this._sendNotification.bind(this));
         $('expand_all').on('click', this._expandAllClicked.bind(this));
 
-        $('show-matches-score').on('input', function() {this._utils.clearHint('score-validation-message');}.bind(this));
-        $('show-matches-phen-score').on('input', function() {this._utils.clearHint('score-validation-message');}.bind(this));
-        $('show-matches-gen-score').on('input', function() {this._utils.clearHint('score-validation-message');}.bind(this));
-
-        $('show-matches-score').on('change', function() {this.validateScore($('show-matches-score').value, 'show-matches-score', 'score-validation-message', true);}.bind(this));
-        $('show-matches-phen-score').on('change', function() {this.validateScore($('show-matches-phen-score').value, 'show-matches-phen-score', 'score-validation-message', true);}.bind(this));
-        $('show-matches-gen-score').on('change', function() {this.validateScore($('show-matches-gen-score').value, 'show-matches-gen-score', 'score-validation-message', true);}.bind(this));
+        this._overallScoreSlider = this._initiateScoreSlider('show-matches-score', (!this._isAdmin) ? 0.4 : 0.1, 0.5);
+        this._phenScoreSlider = this._initiateScoreSlider('show-matches-phen-score', 0, 0);
+        this._genScoreSlider = this._initiateScoreSlider('show-matches-gen-score', 0, 0.1);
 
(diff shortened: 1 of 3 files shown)


allasm:
I think it is less convenient now. The table displays the score, while the sliders have some unspecified scale. I think we need to add the number to the slider somehow (to the left/right? on top of the slider handle?)

(I would also remove the shadow, it drags to much attention to itself - while those sliders are not needed for most of the interaction time)

sashaandjic:
I think switching to percentages in filters and in the grid would make more sense to a typical person.

allasm:
Note: overall score is set correctly - the range does not start from 0

I'm not sure percentages make sense. A percent score will rei force the notion of "this much similar" while in fact this is a non-linear score, and e.g genetic score of 0.2 does not mean 20% similar, it doesnt mean anything in particular, just our scoring  algorithm decided that this is the score, amd higher score generally correlates with more similar patients. The only scores that make sense is 0 (no similarity) and 1 (the same - though not necessarilg to be precise)

sashaandjic:
I think the default are still not correct.
screen shot 2018-11-20 at 9 06 00 am
screen shot 2018-11-20 at 9 06 28 am

marta-:
I haven't tested this myself, but if the last screenshot from Sasha is still accurate please find a place to display the selected value (real time) for each slider

veronikaslc:
The last commit is work in progress, please do not test.

veronikaslc:
Ready for test and review.

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -114,6 +105,31 @@ var PhenoTips = (function (PhenoTips) {
         Event.observe(window, 'resize', this._buildTable.bind(this));
     },
 
+    _initiateScoreSlider: function(id, minScore, initialScore) {
```
marta-: Please rename to `initializeScoreSlider`

veronikaslc:
TODO: fix the slider handle relative positioning (ex. when the slider says .5 but doesn't look like it is at .5 on the scale)

marta-:
The positioning of the handle is still corresponding to the actual value as can be seen in the screenshots:

image

image

The middle of the slider should always be on the value. That means the side wil somethimes be outside the scale line:
image

Setting the margin left to be negative half of the width and allowing the handle to move until its center reaches the end of the scale should fix this.

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -119,6 +110,32 @@ var PhenoTips = (function (PhenoTips) {
...
+        var progressEl = filterEl.down('.progress');
+        var handleEl = filterEl.down('.handle');
+
+        var slider = new Control.Slider(handleEl, filterEl, {
+            range: $R(minScore, 1),
+            sliderValue: initialScore,
+            onSlide: function(value) {
+                progressEl.setStyle({ width: parseInt(window.getComputedStyle(handleEl, null).getPropertyValue("left").replace('px', '')) + 2 + "px"});
+                // round a float to the nearest 0.05
+                handleEl.update((Math.ceil(value*20)/20).toFixed(2));
```
marta-: Precision of 2 decimals doesn't look great for numbers like `0` or `0.5`, especially when you move a slider that is at its initial value just a little bit, and it turns from `0` to `0.00`. This may help:  
```suggestion
                handleEl.update((Math.ceil(value*20)/20).toFixed(2).toString().replace(/[\.]?0+$/, ''));
```

veronikaslc:
Addressed last codereview comments, please, see on http://172.20.4.87:8080/MatchingNotification/MatchesTable

marta-:
Brittney finds the fact that one of the sliders starts at 0.1 (or 0.4) and the other 2 start at 0 very confusing. She would like us to reconsider having all sliders start at 0 and somehow disable the range that is not permitted for the avg score.

PR #220, opened 2018-11-21
PN-346: Final modifications for 1.1.13

veronikaslc:


allasm:
A few issues are found:
- if a case is solved but has no pubmedID contact button is now shown
- secondary ("grey") version of contact button is not show if case is already contacted

veronikaslc:
Addressed:
-if a case is solved but has no Pubmed ID contact button is not shown
- secondary ("grey") version of contact button is not show if a match is already contacted (PN-369?)
- in admin table there is an extra (last) column in the header of the table (probably PN-370)

PR #224, opened 2018-11-30
PN-390: My patient link is broken in the contact email template

veronikaslc:
Changed back changes for stable aimed for PN 1.2


allasm:
I searched the entire PN repository for `getDocumentReference()` and there are a lot of matches. Should we review every single one of them to avoid a similar bug elsewhere?

PR #234, opened 2018-12-06
PN-375: Support for notification history

veronikaslc:
Ready for codereview


sashaandjic:
Should I, as a user, be able to see all notifications and contacts sent to me?
For example, if you look at the P0000002 - P0000018 match on http://172.20.4.87:8080/ you can see that both case owners were notified several times by administrators and that Mary contacted John. When I login as Mary I can only see John's notifications and contacts. I cannot see that I was notified 3 times by adminstrators about the same match. Is this correct?

veronikaslc:
Yes, this is correct and implemented according to specifications in the PN-375 jira issue `Add history icon to each of the `Contact emails` columns for admins (for the matched email column for users)`. Matched email column for users means the column for the patient that Mary sends notifications to, in this case, she sends notifications to John.

sashaandjic:
We should rename "contacted" to "notified" when logged as an administrator.
image

sashaandjic:
One more thing: I can mark as contacted only matching pairs that were not notified. The CONTACTED check box is unavailable to Mary in the attached screenshot.
image

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -1108,30 +1271,16 @@ var PhenoTips = (function (PhenoTips) {
...
-    _getNotified : function(match)
-    {
-        var td = '<td style="text-align: center" name="notified">';
-        td += this._getMarkNotifiedButtonHTML(match);
-        td += '</td>';
-        return td;
-    },
-
-    _getMarkNotifiedButtonHTML : function(match)
+    _getMarkUserContactedHTML : function(match)
```
allasm: "getMark____" is a very confusing word combination, something like `getManualContactButtonHTML()` would be better

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -1007,21 +1065,123 @@ var PhenoTips = (function (PhenoTips) {
...
 
         //if logged as admin - add notification checkbox for local PC patient email contact but not for self (not for patients owned by admin)
-        if (this._isAdmin && serverId == '' && emails.length > 0) {
-            td += '<span class="fa fa-envelope" title="' + this._NOTIFY + '"></span> <input type="checkbox" class="notify" data-matchid="' + matchId + '" data-patientid="'+ patientId +'" data-emails="'+ emails.toString() +'">';
+        if (this._isAdmin && patient.serverId == '' && patient.emails.length > 0) {
+            td += '<span class="fa fa-envelope" title="' + this._NOTIFY + '"></span> <input type="checkbox" class="notify" data-matchid="' + matchId + '" data-patientid="'+ patient.patientId +'" data-emails="'+ patient.emails.toString() +'">';
         }
         td += '</td>';
         return td;
     },
```
allasm: Note sure where in the code the change should be, but when a match is contacted outside of PC, the button "contact" is greyed out, but the history icon is also grey, which is a bit strange - greyed out button means it was contacted, but supposedly there is still no history

I would also make the "on hover" CSS for the history button more noticeable, it is absolutely not clear that the greyed out button can still be pressed (well, the cursor does change shape, but still)

Also I would add a bit more space between the button and the history icon, and decrease space on the right

allasm:
I think we also already have "contacted outside of PC" button in production, we need to check if there are any records and migrate those to new history format if there are any

matching-notification-api/src/main/java/org/phenotips/matchingnotification/notification/internal/AbstractPatientMatchEmail.java
```diff
@@ -452,4 +472,26 @@ public JSONObject getEmail()
 
         return result;
     }
+
+    private JSONObject getPatientContactInfo(Patient patient)
+    {
+        if (patient != null) {
+            PatientData<ContactInfo> data = patient.getData("contact");
+            if (data != null && data.size() > 0) {
+                ContactInfo contact = data.get(0);
```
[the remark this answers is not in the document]
veronikaslc: origins from [here](https://github.com/phenotips/remote-matching/commit/856017bbb28a03b14c3f79d72366bbb8106afbb4#diff-decf87697fbc39d0321619a18ac71232R121)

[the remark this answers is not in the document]
allasm: Hmm, I don't see any code there that behaves this way?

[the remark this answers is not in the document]
allasm: Well, that is code in remote matching repository which tries to collect data to fit into the MME protocol, which is not the restriction we have here. I don't think it is critical to fix this right now, but I think this is wrong and should be fixed later

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
veronikaslc: I'd leave it like that

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -1007,21 +1065,123 @@ var PhenoTips = (function (PhenoTips) {
...
 
-    _getContact : function(record)
+    _generateNotificationHistoryTable : function(records)
+    {
+        var table = '<table class="notification-history-table"><tbody>';
```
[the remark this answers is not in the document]
veronikaslc: yes, we can, may be in the next release

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -1007,21 +1065,123 @@ var PhenoTips = (function (PhenoTips) {
...
+        if (type == "notification" && info.userinfo.id == "xwiki:XWiki.Admin") {
+            return "PhenomeCentral";
+        }
+
+        var cell = '';
+        // user name and institution
+        if (info.userinfo) {
+            if (info.userinfo.name) {
+                if (info.userinfo.id) {
+                    var space = (info.userinfo.id.startsWith("xwiki:XWiki.")) ? "XWiki" : "Groups";
```
marta-: This is brittle. Please either don't link to tye user's profile or send the url in the JSON instead of generating it here.

sashaandjic: This will not work for Brittney nor any other admin 
image

I believe it is causing problem shown below  where Peter is an administrator sending notification (I will log it separatelly)
image

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -1007,21 +1065,123 @@ var PhenoTips = (function (PhenoTips) {
...
+
+        table += '</tbody></table>';
+
+        return table;
+    },
+
+    _generateRecipientCell : function(info, type)
+    {
+        // notifications should have something generic e.g. "PhenomeCentral" (no link)
+        if (type == "notification" && info.userinfo.id == "xwiki:XWiki.Admin") {
```
marta-: >This will not work for Brittney nor any other admin

Right, good catch @sashaandjic . The `From` field should be " PhenomeCentral" for all `notification`s, regardless of which admin sent them.

veronikaslc:
I think 172.20.4.128 build got broken, weird error in instance logs:
`java.net.URISyntaxException: Illegal character in query at index 98: http://172.20.4.128:8080/bin/ServerError?netnum=127.0.0.1&netmask=255.255.255.0&netid=127.0.0.1%20|%20id|`

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -1124,10 +1124,10 @@ var PhenoTips = (function (PhenoTips) {
         return table;
     },
 
-    _generateRecipientCell : function(info, type)
+    _generateRecipientCell : function(info, type, isFrom)
```
marta-: If `isFrom`, then this is not really a `recipient` cell. The name of the function is inaccurate and confusing. Can we call it something more generic such as `_generateUserInfoCell`?

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -1124,10 +1124,10 @@ var PhenoTips = (function (PhenoTips) {
...
     {
-        // notifications should have something generic e.g. "PhenomeCentral" (no link)
-        if (type == "notification") {
+        // The `From` field should be " PhenomeCentral" for all `notification`s, regardless of which admin sent them
+        if (type == "notification" && isFrom) {
```
marta-: For better code readability, I would actually move the `if` up to line 1104, and only pass 1 parameter `info` to this function.

marta-: Another comment about lines 1136-1145 below: the code for obtaining the URL is brittle. - please generate the URL in the backend and include it in the JSON with the rest of the user info.

veronikaslc: @marta- should we also change it [here](https://github.com/phenotips/phenotips/blob/master/components/patient-access-rules/ui/src/main/resources/PhenoTips/AccessRightsManagement.xml#L1013)?

matching-notification-api/src/main/java/org/phenotips/matchingnotification/notification/internal/AbstractPatientMatchEmail.java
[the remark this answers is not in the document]
veronikaslc: @allasm some details to the story. The reason we have more than one `ContactInfo` is that `ContactInformationController` in loops through all `PatientContactProvider`s and collects their info. Currently we have two: `OwnerContactProvider` (priority 100) and `ReporterContactProvider` (priority 900). `data.get(0)` returns reporter info. Do we need them all?

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
marta-: ^ Yes, please create an issue for it.

sashaandjic:
ui changes:
- before opening notification history close any notification history that is open 
- remove sender's link 
- change tooltips "contact" to "Email sent"; "notification" to "notification sent"
- make sure above tooltips are shown when howering over column where the icon & date are shown only (not for the whole row)

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
veronikaslc: Created [PT-3942](https://phenotips.atlassian.net/browse/PT-3942) for it.

PR #238, opened 2019-01-11
PN-404: Add ability to record "My notes" to each matching pair

veronikaslc:
Done.

### DIFF
--- a/matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
+++ b/matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
@@ -546,7 +546,7 @@ public class DefaultMatchStorageManager implements MatchStorageManager
     }
 
     @Override
-    public boolean saveNote(List<PatientMatch> matches, String note)
+    public boolean addNote(List<PatientMatch> matches, String note)
     {
         Session session = this.beginTransaction();
         boolean transactionCompleted = false;
--- a/matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
+++ b/matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
@@ -182,6 +182,7 @@ public class DefaultMatchStorageManager implements MatchStorageManager
                         match.setStatus(existingMatch.getStatus());
                         match.setComment(existingMatch.getComment());
                         match.setNotificationHistory(existingMatch.getNotificationHistory());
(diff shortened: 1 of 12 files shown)


matching-notification-api/src/main/resources/ApplicationResources.properties
```diff
@@ -68,6 +68,9 @@ phenotips.matchingNotifications.table.notify=Mark match contact to notify
 phenotips.matchingNotifications.table.notify.noContactSelected=No contacts selected for notifying
 phenotips.matchingNotifications.table.saveComment=Save comment
 phenotips.matchingNotifications.table.addComment=Add comment
+phenotips.matchingNotifications.table.notes.title=My notes
+phenotips.matchingNotifications.table.notes.save=Save
+phenotips.matchingNotifications.table.notes.hint=Only you can see those notes
```
marta-: "those" -> "these"

matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
```diff
@@ -180,6 +180,7 @@ public boolean saveLocalMatches(List<PatientMatch> matches, String patientId)
                         match.setFoundTimestamp(existingMatch.getFoundTimestamp());
                         match.setComment(existingMatch.getComment());
                         match.setNotificationHistory(existingMatch.getNotificationHistory());
+                        match.setNotes(existingMatch.getNotes());
```
allasm: At this point I wonder if we should remove all setXXX() methods (which are only ever used here, and may be misleading to programmer-users since everywhere else updateXXX() methods should be used) and have one clone() or transfer() method which takes care of this particular use case

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/MatchingNotificationResource.java
```diff
@@ -226,4 +226,17 @@ Response setUserContacted(@FormParam("matchesIds") Set<Long> matchesIds,
     @Produces(MediaType.APPLICATION_JSON)
     @Path("/save-comment")
     Response saveComment(@FormParam("matchesIds") Set<Long> matchesIds, @FormParam("comment") String comment);
+
+    /**
+     * Saves note for matches, with ids given in parameter.
```
allasm: I would rename all public `saveNote()` methods to `addNote()`, to clearly indicate that notes are never edited or replaced, and only new ones are added (by design) - would add this to the comments as well

veronikaslc: renamed

PR #239, opened 2019-01-15
PN-376: Preserve comment history per contributing user

veronikaslc:


sashaandjic:
Current user's comments are not saved and are always opened for editing. 
Login as Admin on http://172.20.4.116:8080/ and go to the Matching Notifications page and set notified and not notified flags on. Observe first 3 matching pairs. All Admins comments are editable.
If you log as Pat, all Pat's comments will be editable to her. 

Also, notice the Notes icon weird positioning in the attached screenshot.
image

sashaandjic:
Everything looks great, except in one case. 

Go to http://172.20.4.133:8080
Log as Julia; type "8" in PatientID contains
Observe the following matches:
julia
Match P8 - P14: Julia clicked on the Contacted checkbox (no email sent)
Match P8 - P15: Julia clicked on orange Contact button (email sent to Pat)

Log out
Login as Pat; type "8" in PatientID contains
Notice that P8-P15 match does not have history and the Contact button is orange. Should it be this way? Julia contacted Pat via email.
pat

veronikaslc:
@sashaandjic Short answer to all questions - yes, it should be this way.
As stated in the description of JIRA issue PN-399: 

> Show Contact button as secondary only **if matched side has a 'contact' type record** in the notification history or a match was contacted by the user outside the PhenomeCentral

As stated in JIRA issue PN-375: 

> Add history icon to each of the `Contact emails` columns for admins (**for the matched email column for users**)

matching-notification-migrations/src/main/java/org/phenotips/matchingnotification/internal/R74691PatientNetwork377DataMigration.java
```diff
@@ -0,0 +1,114 @@
...
+        Session session = this.sessionFactory.getSessionFactory().openSession();
+        Transaction t = session.beginTransaction();
+
+        try {
+            SQLQuery changeName = session.createSQLQuery(
+                "ALTER TABLE patient_matching CHANGE comment comments VARCHAR(0xFFFFFF)");
+            changeName.executeUpdate();
+
+            Query q1 = session.createQuery("update " + PatientMatch.class.getCanonicalName()
+                + " set comments = ('{ comments:[{ comment:' + comments + '}] }') where comments is not null");
```
sdumitriu: `set comments = concat('{ comments:[{ comment:', comments, '}] }') where...`

allasm:
+1 to Sasha's comments. Clicking on comment icons below the first screen jumps away effectively making it unusable. Also that jump replaces all black "have comments" icons with grey "no comments" icons.

Also clicking on user link in the comments opens a page with an incorrect URL

veronikaslc:
comments addressed

sashaandjic:
The text below seems to be taking too much space. Should we put it under "i" button? @marta- 
Any user, collaborator or group associated with either case in this match can see and post in discussion. Once saved, comments cannot be edited.
image

allasm:
@sashaandjic text is OK, we can make the dialogue a bit wider to reduce the number of lines by one.
A larger problem is that the text is not correct, since not "any user" can see it, but only admins

sashaandjic:
Agree. It's cosmetic complain. It kind of dominates visually. Would smaller italicized font help?

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -870,11 +881,19 @@ var PhenoTips = (function (PhenoTips) {
...
-            td += '<div class="xTooltip comment-container"><span class="hide-tool" title="Hide">×</span><div><textarea rows="5" cols="20"></textarea></div>'
-                +'<span class="buttonwrapper"><a class="button save-comment" data-matchid="' + match.id + '" href="#"><span class="fa fa-save"> </span>'
-                + this._SAVE_COMMENT_BUTTON_LABEL + '</a></span></div>';
+            var commentsTable = '';
+            if (match.comments) {
+                commentsTable = this._generateCommentsTable(match.comments);
+            }
+            td += '<div class="xTooltip comment-container"><span class="hide-tool" title="Hide">×</span>'
+                + '<div class="nhdialog-title">' + this._COMMENTS_TITLE + '</div>'
+                + '<p class="xHint">' + this._COMMENTS_HINT + '</p>'
```
marta-: Didn't we decide yesterday to turn this into an (i) or (?) button?

veronikaslc: We were testing yesterday with Sasha and noticed that notes have this info not styled as icon, so for now for consistency with notes dialog we can leave like text for now and change it later. I made it smaller with less line space so it is not taking much space now.

veronikaslc: We can move text to help icons when we make all pop-ups as a separate widget later.

veronikaslc:
it's expected behaviour, we return 400 Bad Request response and nothing changes in UI.

sashaandjic:
> a very small thing... if you hit SAVE bttn on an either empty or blank Note, or ADD bttn on a blank comment, this happens
> image

Only this happens in ui 
image

matching-notification-migrations/src/main/java/org/phenotips/matchingnotification/internal/R74693PatientNetwork377DataMigration.java
```diff
@@ -0,0 +1,106 @@
...
+import com.xpn.xwiki.store.migration.DataMigrationException;
+import com.xpn.xwiki.store.migration.XWikiDBVersion;
+import com.xpn.xwiki.store.migration.hibernate.AbstractHibernateDataMigration;
+
+/**
+ * Migration for PatientNetwork issue PN-377: Migrate existing comments to the new JSON string format to store comments
+ * from multiple users. Existing comments will be migrated to be contained in the JSON object.
+ *
+ * @version $Id$
+ * @since 1.3
```
sdumitriu: 1.2?

matching-notification-migrations/src/main/java/org/phenotips/matchingnotification/internal/R74693PatientNetwork377DataMigration.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @Override
+    public void hibernateMigrate() throws DataMigrationException, XWikiException
+    {
+        Session session = this.sessionFactory.getSessionFactory().openSession();
+        Transaction t = session.beginTransaction();
+
+        try {
+            Query q1 = session.createQuery("update " + PatientMatch.class.getCanonicalName()
+                + " set comments = concat('[{ comment:', comment, '}]') where comment <> ''");
```
sdumitriu: I see that a comment object has a lot more data when set through the api. Should this simple code be replaced with a more complicated java update, which also sets thr user and date? If not, will any other code break if this info is missing?

veronikaslc: We do not know user or date here, so no correct way to set it. If user name or date are not present, in UI we just do not show it.

PR #240, opened 2019-01-15
PN-416: Solved cases in Pre-Prod Match Table list blank PMID links rather than contact e-mails

veronikaslc:
exclude empty/blank pubmedIDs from displaying


veronikaslc:
Closed, will be fixed in PR #[2807](https://github.com/phenotips/phenotips/pull/2807).

PR #241, opened 2019-01-16
PN-417: Display all match locations in matching table with corresponding number of matches

veronikaslc:
Done.

### DIFF
--- a/matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
+++ b/matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
@@ -87,8 +87,6 @@ var PhenoTips = (function (PhenoTips) {
         // set behaviour for hide/show email columns triggers
         this._initiateEmailColumnsBehaviur();
 
-        $('checkbox-server-filters').hide();
-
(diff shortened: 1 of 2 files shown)


matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -1183,14 +1170,13 @@ var PhenoTips = (function (PhenoTips) {
...
     {
-        if (this._presentServerIds.length > 1) {
-            $('checkbox-server-filters').show();
-            $('matching-filters').select('input[type="checkbox"][name="checkbox-server-id-filter"]').each( function(selectEl) {
-                (this._presentServerIds.indexOf(selectEl.value) >= 0) ? selectEl.up().show() : selectEl.up().hide();
-            }.bind(this));
-        } else {
-            $('checkbox-server-filters').hide();
-        }
+        $('checkbox-server-filters').show();
```
marta-: This line should be removed, together with line 90 which hides the element in the first place `$('checkbox-server-filters').hide()` since they just cancel each other out.

sashaandjic:
Maybe smaller fonts would look better? Check it on  http://172.20.4.139:8080/
image

PR #243, opened 2019-02-04
PN-420: When user has more than one email contact dialog is broken

veronikaslc:
Fixed parsing user emails fetched from the user profile

### DIFF
--- a/matching-notification-api/src/main/java/org/phenotips/matchingnotification/notification/internal/DefaultUserPatientMatchEmail.java
+++ b/matching-notification-api/src/main/java/org/phenotips/matchingnotification/notification/internal/DefaultUserPatientMatchEmail.java
@@ -132,9 +132,7 @@ public class DefaultUserPatientMatchEmail extends AbstractPatientMatchEmail
         try {
             Set<Address> userEmails = getUserEmails();
             for (Address email : userEmails) {
-                if (email != null) {
-                    this.mimeMessage.addRecipient(RecipientType.CC, email);
-                }
+                this.mimeMessage.addRecipient(RecipientType.CC, email);
             }
         } catch (Exception ex) {
             // do nothing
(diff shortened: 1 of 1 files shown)


matching-notification-api/src/main/java/org/phenotips/matchingnotification/notification/internal/DefaultUserPatientMatchEmail.java
```diff
@@ -142,24 +146,31 @@ protected void setFrom()
     {
         super.setFrom();
         try {
-            Address userEmails = getUserEmails();
+            Set<Address> userEmails = getUserEmails();
             if (userEmails != null) {
```
marta-: `getUserEmails()` is a private method that always returns a non-null result. There's no need to keep this null check.

matching-notification-api/src/main/java/org/phenotips/matchingnotification/notification/internal/DefaultUserPatientMatchEmail.java
```diff
@@ -128,9 +130,11 @@ protected void setTo()
     {
         super.setTo();
         try {
-            Address userEmail = getUserEmails();
-            if (userEmail != null) {
-                this.mimeMessage.addRecipient(RecipientType.CC, userEmail);
+            Set<Address> userEmails = getUserEmails();
+            for (Address email : userEmails) {
+                if (email != null) {
```
marta-: `getUserEmails()` should ensure that no null objects are returned so that this null check is unnecessary.

matching-notification-api/src/main/java/org/phenotips/matchingnotification/notification/internal/DefaultUserPatientMatchEmail.java
```diff
@@ -142,24 +146,31 @@ protected void setFrom()
...
+    private Set<Address> getUserEmails()
     {
         String userEmail = USERMANAGER.getCurrentUser().getAttribute(USER_PROPERTY_EMAIL).toString();
-        if (StringUtils.isNotBlank(userEmail)) {
-            InternetAddress replyTo = new InternetAddress();
-            replyTo.setAddress(userEmail);
-            return replyTo;
+        Set<Address> emails = new HashSet<>();
+
+        for (String parsedEmail : StringUtils.split(userEmail, ",|;")) {
```
marta-: Should space be one of the separators? Would anyone list multiple emails like this?
```
email1@domain1.com email2@domain2.com
```

I think it's much more likely to use ` ` than `|` as a separator.

PR #244, opened 2019-02-05
PN-420-stable-1.x: When user has more than one email contact dialog is broken

veronikaslc:
Fixed


matching-notification-api/src/main/java/org/phenotips/matchingnotification/notification/internal/DefaultUserPatientMatchEmail.java
```diff
@@ -142,24 +144,29 @@ protected void setFrom()
...
+    private Set<Address> getUserEmails()
     {
         String userEmail = USERMANAGER.getCurrentUser().getAttribute(USER_PROPERTY_EMAIL).toString();
-        if (StringUtils.isNotBlank(userEmail)) {
-            InternetAddress replyTo = new InternetAddress();
-            replyTo.setAddress(userEmail);
-            return replyTo;
+        Set<Address> emails = new HashSet<>();
+
+        for (String parsedEmail : StringUtils.split(userEmail, ",|;| ")) {
```
sdumitriu: This is not a regexp, use `"|,; "` instead.

PR #247, opened 2019-02-19
PN-425: Better usability of send notifications button

veronikaslc:
Prevent jumping to the top of the Matching notification page after clicking on SEND NOTIFICATION button
Hide SEND NOTIFICATION button if table is empty
Feedback from sending notification should be close to the button

### DIFF
--- a/matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
+++ b/matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
@@ -36,8 +36,9 @@ var PhenoTips = (function (PhenoTips) {
         this._utils = new utils(this._tableElement);
 
         $('show-matches-button').on('click', this._showMatches.bind(this));
-        $('send-notifications-button').addClassName("disabled");
-        $('send-notifications-button').on('click', this._sendNotification.bind(this));
+        this._notificationsButton = $('send-notifications-button');
+        this._notificationsButton.hide();
+        this._notificationsButton.on('click', this._sendNotification.bind(this));
         $('expand_all').on('click', this._expandAllClicked.bind(this));
 
         this._overallScoreSlider = this._initializeScoreSlider('show-matches-score', (!this._isAdmin) ? 0.4 : 0.1, 0.5);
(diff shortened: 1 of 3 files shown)


matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -486,7 +487,7 @@ var PhenoTips = (function (PhenoTips) {
             onCreate : function () {
                 $("panels-livetable-ajax-loader").show();
                 this._utils.clearHint('send-notifications-messages');
-                $('send-notifications-button').addClassName("disabled");
```
marta-: I'm not sure why `send-notification-button` is implemented as a `<a>` element. IMO it should really be a button, and that would make many things easier:
* Jumping to the top of the document when pressing "Send notifications" would never happen, no need to stop the event to prevent it
* It can be `.disable()`d for real , not just styled as disabled, preventing any functionality occurring when there's no notification to send

marta-:
> SEND NOTIFICATIONS button is not hidden when the table is empty. It is disabled, but user can click on it.
> image

This problem stems from a bug that is not tied to this PR. It affects production as well and has been reported as [PN-436](https://phenotips.atlassian.net/browse/PN-436).

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
veronikaslc: I was thinking, that in a way it might be useful to give a feedback to user when there is no emails selected and user still can click on the button and see this message as a hint on what to do to use this functionality. So thats why I left the button like this, but may be its not that needed.

marta-: In my opinion, the fact that the button is disabled should be enough in this case. Also, this is the admin UI, not for broad usage, so we don't need that many explanations. Robust and easily maintainable code logic is more important IMO.

veronikaslc:
Should the button be completely hidden when user filters out all loaded matches? I am not strongly opinionated, but I lean towards the still showing button styled as a secondary since there are some loaded matches.

marta-:
> Should the button be completely hidden when user filters out all loaded matches? I am not strongly opinionated, but I lean towards the still showing button styled as a secondary since there are some loaded matches.

I don't have a strong opinion either, as long as it's clear for the user what the state of the system is.

PR #248, opened 2019-02-20
PN-431: API for retrieving "My Matches"

veronikaslc:
Added fromDate and toDate filters
Removed unused onlyNotified from API parameters for retrieving matches

### DIFF
--- a/matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/MatchStorageManager.java
+++ b/matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/MatchStorageManager.java
@@ -41,8 +41,10 @@ public interface MatchStorageManager
      * @param phenScore only matches with phenotypical score higher or equal to this value are returned
      * @param genScore only matches with genotypical score higher or equal to this value are returned
      * @param onlyCurrentUserAccessible when true only matches that current user has access to are returned
-     * @param fromDate only matches found on or after the specified date (0 by default)
-     * @param toDate only matches found on or before the specified date (current date by default)
+     * @param fromDate if passed a date in the {@code yyyy/MM/dd} format, then only matches found on or after this date
+     *        will be returned; if {@code null} or an empty string, then no lower limit on the match date is considered
+     * @param toDate if passed a date in the {@code yyyy/MM/dd} format, then only matches found on or before this date
+     *        will be returned; if {@code null} or an empty string, then no upper limit on the match date is considered
      * @return a list of matches
      */
     List<PatientMatch> loadMatches(double score, double phenScore, double genScore, boolean onlyCurrentUserAccessible,
--- a/matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/MatchStorageManager.java
+++ b/matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/MatchStorageManager.java
@@ -41,9 +41,12 @@ public interface MatchStorageManager
      * @param phenScore only matches with phenotypical score higher or equal to this value are returned
      * @param genScore only matches with genotypical score higher or equal to this value are returned
      * @param onlyCurrentUserAccessible when true only matches that current user has access to are returned
+     * @param fromDate only matches found on or after the specified date (0 by default)
+     * @param toDate only matches found on or before the specified date (current date by default)
      * @return a list of matches
      */
-    List<PatientMatch> loadMatches(double score, double phenScore, double genScore, boolean onlyCurrentUserAccessible);
+    List<PatientMatch> loadMatches(double score, double phenScore, double genScore, boolean onlyCurrentUserAccessible,
+        String fromDate, String toDate);
 
     /**
      * Load all matches with ids in {@code matchesIds}.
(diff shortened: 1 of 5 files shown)


matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/MatchStorageManager.java
```diff
@@ -41,9 +41,12 @@
      * @param phenScore only matches with phenotypical score higher or equal to this value are returned
      * @param genScore only matches with genotypical score higher or equal to this value are returned
      * @param onlyCurrentUserAccessible when true only matches that current user has access to are returned
+     * @param fromDate only matches found on or after the specified date (0 by default)
+     * @param toDate only matches found on or before the specified date (current date by default)
```
sdumitriu: These are not correct, and not informative enough. A `fromDate` of `0` doesn't make sense, since there's no such date, and there's actually no default value for them, since Java doesn't have default values for parameters. There is a default behavior for `null` or empty strings. And what is the expected format of the date?

```
     * @param fromDate if passed a date in the {@code yyyy/MM/dd} format, then only matches found on or after this date will be returned; if {@code} null or the empty string, then no lower limit on the match date is considered
     * @param toDate if passed a date in the {@code yyyy/MM/dd} format, then only matches found on or before this date will be returned; if {@code} null or the empty string, then no upper limit on the match date is considered
```

And even this is a bit incorrect, because the timestamp parsed from this date refers to the start of the date, so technically the `toDate` will match events occurring exactly at midnight, not later during the day, so the behavior isn't "on or before", but just "before".

allasm:
@sashaandjic To test API you can use the following CURL: 

`curl -u Admin:admin -H "Content-Type: application/json" -X POST -d 'score=0.50&phenScore=0.00&genScore=0.10&toDate=2019%2F02%2F21&fromDate=2019%2F02%2F21' http://172.20.4.126:8080/rest/patients/matching-notification/show-matches`

Note that dates are encoded with `%2F` replacing `/`, e.g `2019/02/21` becomes `2019%2F02%2F21`

allasm:
tested, seems OK

matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/MatchStorageManager.java
allasm: > because the timestamp parsed from this date refers to the start of the date

I think this is an important point, what do you think about adding 24hrs to the timestamp of the provided `toDate`?

matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
```diff
@@ -285,23 +285,16 @@ public boolean saveRemoteMatches(List<? extends PatientSimilarityView> similarit
...
-            if (StringUtils.isNotBlank(toDate)) {
-                try {
-                    Timestamp timestampTo = new Timestamp(sdf.parse(toDate).getTime());
-                    query.setTimestamp("toTimestamp", timestampTo);
-                } catch (Exception e) {
-                    // do nothing
-                }
+            if (toDate != null) {
+                // need to add 1 day in ms to include all matches made that day
+                Timestamp timestampTo = new Timestamp(toDate.getTime() + 86400000);
```
allasm: It is better to write `1000*60*60*24` than `86400000`, then it is clear where the number comes from

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/internal/DefaultMatchingNotificationResource.java
```diff
@@ -116,15 +119,36 @@ public void refreshMatches(Set<String> serverIds, boolean onlyCheckPatientsUpdat
 
     @Override
     public Response getMatches(@Nullable final double score, @Nullable final double phenScore,
-        @Nullable final double genScore, final boolean onlyNotified)
+        @Nullable final double genScore, final String fromDate, final String toDate)
     {
+        SimpleDateFormat sdf = new SimpleDateFormat("yyyy/MM/dd");
+        Date from = null;
+        if (StringUtils.isNotBlank(fromDate)) {
```
allasm: the code could be separated into a `parseDate()` function

matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
```diff
@@ -119,6 +121,8 @@
     private static final String HQL_QUERY_REMOTE_MATCHES =
         "select count(*) from DefaultPatientMatch as m where m.referenceServerId != '' or m.matchedServerId !=''";
 
+    private static final int ONE_HOUR_IN_MILLISECONDS = 1000 * 60 * 60 * 24;
```
marta-: This is one _day_ in milliseconds.

allasm: yes. sorry , that is what we need and what I meant to say

PR #249, opened 2019-02-20
Fixes in match-finding related to switch to PT 1.4

allasm:


similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/DefaultPatientSimilarityViewFactory.java
```diff
@@ -132,8 +139,22 @@ public PatientSimilarityView makeSimilarPatient(Patient match, Patient reference
...
         }
-        AccessType access = new DefaultAccessType(this.permissions.getEntityAccess(match).getAccessLevel(),
-            this.viewAccess, this.matchAccess);
+
+        // FIXME: a patient may have visibility level "matchable" but have access level below "match".
+        //        this does not allow MME code to return matchable patients, and so a workaround is
+        //        implemented below. The workaround should be removed if the condition is no longer possible
+        //        - or the entire system need to be redone to use visibility instead of access
+        EntityAccess entityAccess = this.permissions.getEntityAccess(match);
+        AccessLevel useAccess = entityAccess.getAccessLevel();
```
marta-: usedAccess? accessToUse?

PR #251, opened 2019-02-25
PN-438: Improve match and match history processing code in matchesTable.js

veronikaslc:


sashaandjic:
How can I test this?

veronikaslc:
nothing particular to check, just in general functionality of notes and notification history

veronikaslc:
addressed, VM is ready for retest

marta-:
@sashaandjic Please clear cache. There's a js error if the new markup is used with the old js code.

veronikaslc:
this is a new VM after PR rebase, its empty

BrianLiSK:
Yeah, I got that weird looking error too. Ctrl+ Shift + R a few times for hard refresh to clear cache and the pagination js.

So far, it looks good, the notes stay there too. However, one minor thing (that might be unrelated) is that the green "Sending notification..." text looks a bit misaligned. It used to be right beside the "Send Notifications" button. I'm not sure if this change caused it or it was the pagination and I just missed it.
greentextalignment

marta-:
> the green "Sending notification..." text looks a bit misaligned. It used to be right beside the "Send Notifications" button. I'm not sure if this change caused it or it was the pagination and I just missed it.

It was the pagination: https://github.com/phenotips/patient-network/commit/95aff808c4818ec3e330087362e73ae3584dc709#diff-a8cf22c56ce0ade395eecd68b594fe06R119

veronikaslc:
fixed in master [here](https://github.com/phenotips/patient-network/commit/ee7f9bf52118df50a0bafcb4ae3b85445ee0a399)

matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientMatch.java
```diff
@@ -385,12 +379,11 @@ public JSONArray getComments()
...
     {
         try {
-            JSONObject allNotes = (this.notes != null) ? new JSONObject(this.notes)
-                : new JSONObject();
-            JSONArray records = allNotes.optJSONArray(NOTES);
-            if (records == null) {
+            if (this.notes == null) {
                 return null;
             }
+            JSONArray records = (this.notes != null) ? new JSONArray(this.notes)
```
marta-: `this.notes` is not `null` if we reached this this point. We can get rid of the null check.

PR #252, opened 2019-03-06
PN-441: Show page numbers on both sides of the My Matches grid and shorten list of page numbers

veronikaslc:
Done

### DIFF
--- a/matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matcherPaginator.js
+++ b/matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matcherPaginator.js
@@ -1,47 +1,60 @@
 var PhenoTips = (function(PhenoTips) {
   var widgets = PhenoTips.widgets = PhenoTips.widgets || {};
   widgets.MatcherPaginator = Class.create({
-    initialize: function(table, domNode, max) {
+    initialize: function(table, domNodes, max) {
       var _this = this;
       this.table = table;
       this.max = max;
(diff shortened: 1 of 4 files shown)


BrianLiSK:
Actually, it looks like the navigation arrows have broke. Tested using both Chrome/FF on: http://172.20.4.125:8080/
arrows

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matcherPaginator.js
```diff
@@ -1,47 +1,60 @@
...
+      this.pagesNodes = [];
+      domNodes.each(function(elem){
+          _this.pagesNodes.push(elem.down(".xwiki-livetable-pagination-content"));
       });
-      nextPagination && nextPagination.observe("click", function(ev) {
-        _this.gotoNextPage(ev);
+
+      this.pagesNodes.each(function(elem){
+          elem.down(".prevPagination") && elem.down(".prevPagination").observe("click", function(ev) {
+              _this.gotoPrevPage(ev);
```
BrianLiSK: I don't think this is ever reached.

marta-: This can't work because the arrow elements are not children of the `xwiki-livetable-pagination-content`. You's need to go up to `.xwiki-livetable-pagination` first to get to the arrows.

veronikaslc:
applied last changes to VM, ready for re-test

PR #254, opened 2019-03-06
PN-375: Support for notification history

veronikaslc:


matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/PatientMatch.java
```diff
@@ -341,7 +341,6 @@
      *                  "emails":["patcat@catrescuecentre.kom"]
      *                },
      *         "cc":  ["mary@blueberry.com"],
-     *         "subjectPatientId":"P000005"
```
allasm: I thought we decided to keep this information for potential future use? If lost, in some cases it can't be recovered

veronikaslc: We don't use it now, I'm not sure how useful is it to keep.

PR #257, opened 2019-03-07
PN-445: The response of the matches API should include information regarding how and when the returned results were obtained

veronikaslc:
Done

### DIFF
--- a/matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/internal/DefaultMatchingNotificationResource.java
+++ b/matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/internal/DefaultMatchingNotificationResource.java
@@ -87,6 +87,12 @@ public class DefaultMatchingNotificationResource extends XWikiResource implement
 
     private static final String MATCHID_STRING = "matchId";
 
+    private static final SimpleDateFormat DATE_TIME_SDF = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ");
+
+    private static final SimpleDateFormat DATE_SDF = new SimpleDateFormat("yyyy/MM/dd");
+
+    private static final int ONE_DAY_IN_MILLISECONDS = 1000 * 60 * 60 * 24 - 1;
+
     @Inject
     private PatientMatchExport patientMatchExport;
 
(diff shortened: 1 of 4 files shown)


matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/internal/DefaultMatchingNotificationResource.java
```diff
@@ -343,6 +348,23 @@ private Response getMatchesResponse(@Nullable final double score, @Nullable fina
...
+        params.put("minGenScore", genScore);
+
+        if (fromDate != null) {
+            params.put("fromDate", DATE_TIME_SDF.format(fromDate));
+        }
+
+        if (toDate != null) {
+            // need to add 1 day without the last millisecond in ms
+            // to let user know that we include all matches found that day
+            Timestamp timestampTo = new Timestamp(toDate.getTime() + ONE_DAY_IN_MILLISECONDS);
```
allasm: we should update `toDate` only once, before calling `this.matchStorageManager.loadMatches()`, and then put here the value we pass there. Otherwise it may get out of sync

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/internal/DefaultMatchingNotificationResource.java
```diff
@@ -358,7 +380,8 @@ private JSONObject buildMatchesJSONArray(List<PatientMatch> matches)
         final JSONObject matchesJson = this.patientMatchExport.toJSON(matches);
         return new JSONObject()
             .put(RESULTS_LABEL, matchesJson.optJSONArray("matches"))
-            .put(RETURNED_SIZE_LABEL, matchesJson.optJSONArray("matches").length());
+            .put(RETURNED_SIZE_LABEL, matchesJson.optJSONArray("matches").length())
+            .put("dateGenerated", DATE_TIME_SDF.format(new Date()));
```
allasm: also why do we add `dateGenerated` here, but `parameters` on line 367 - makes sense to have all new changes in one place?

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/internal/DefaultMatchingNotificationResource.java
veronikaslc: fair

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/internal/DefaultMatchingNotificationResource.java
```diff
@@ -87,6 +87,12 @@
 
     private static final String MATCHID_STRING = "matchId";
 
+    private static final SimpleDateFormat DATE_TIME_SDF = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ");
+
+    private static final SimpleDateFormat DATE_SDF = new SimpleDateFormat("yyyy/MM/dd");
```
marta-: I should have noticed and commented on PR #248 which added `fromDate` and `toDate`, but we should probably accept `yyyy-MM-dd` for consistency.

veronikaslc:
@marta- codereview comments addressed

PR #259, opened 2019-03-18
Initial implementation of a stand-alone phenotype similarity scoring service

allasm:
Do not merge, this PR is for convenience of monitoring changes only

### DIFF
--- a/matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/MatchingNotificationResource.java
+++ b/matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/MatchingNotificationResource.java
@@ -239,4 +239,18 @@ public interface MatchingNotificationResource
     @Produces(MediaType.APPLICATION_JSON)
     @Path("/save-note")
     Response addNote(@FormParam("matchesIds") Set<Long> matchesIds, @FormParam("note") String note);
+
+    /**
+     * Computes phenotype similarity score.
+     *
+     * @param referencePatient JSON
+     * @param matchPatient JSON
+     * @return result JSON
+     */
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/compute-score")
+    Response computeScore(@FormParam("referencePatient") String referencePatient,
+            @FormParam("matchPatient") String matchPatient);
 }
--- a/matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/internal/DefaultMatchingNotificationResource.java
+++ b/matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/internal/DefaultMatchingNotificationResource.java
@@ -17,6 +17,9 @@
  */
 package org.phenotips.matchingnotification.rest.internal;
 
+import org.phenotips.data.Patient;
+import org.phenotips.data.PatientRepository;
+import org.phenotips.data.similarity.PatientSimilarityScorer;
 import org.phenotips.matchingnotification.MatchingNotificationManager;
 import org.phenotips.matchingnotification.export.PatientMatchExport;
 import org.phenotips.matchingnotification.finder.MatchFinderManager;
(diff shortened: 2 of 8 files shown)


matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/MatchingNotificationResource.java
```diff
@@ -239,4 +239,18 @@ Response setUserContacted(@FormParam("matchesIds") Set<Long> matchesIds,
...
+
+    /**
+     * Computes phenotype similarity score.
+     *
+     * @param referencePatient JSON
+     * @param matchPatient JSON
+     * @return result JSON
+     */
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
```
sdumitriu: This doesn't really consume json, since it manually parses jsons from an x-www-form-encoded list of parameters.

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/internal/DefaultMatchingNotificationResource.java
```diff
@@ -482,4 +491,39 @@ private Long getLastOutgoingMatchId(List<PatientMatch> matches)
...
+    private Patient createPatientFromJSON(String patientJSON)
+    {
+        JSONObject json;
+        try {
+            json = new JSONObject(patientJSON);
+        } catch (Exception ex) {
+            this.slf4Jlogger.error("Can't convert JSON to a patient: [{}]", ex.getMessage());
+            return null;
+        }
+        final Patient patient = this.repository.create();
```
sdumitriu: This actually creates a patient, so it is not suitable for use in production.

Some alternatives:
- Make sure you delete the patient at the end (preferably skipping the document trash), although this will create some churn
- Although using internal classes is discouraged, in this case it would be better to manually create a PhenoTipsPatient on a new XWikiDocument in a temporary space, and later delete that document
- Implement the `Patient` interface like it's done in `org.phenotips.data.script.PatientSpecificityScriptService.FakePatient`, with no actual persistence and only implementing the methods the score cares about
- Make that an actual public implementation, `TransientPatientImpl` that other services can use for temporary storage; it could extend `PhenoTipsPatient` and override methods so that they never actually save the underlying `XWikiDocument`, but otherwise it would behave just like a real patient

similarity-data-impl/pom.xml
```diff
@@ -27,7 +27,7 @@
   </parent>
 
   <properties>
-    <coverage.instructionRatio>0.59</coverage.instructionRatio>
+    <coverage.instructionRatio>0.54</coverage.instructionRatio>
```
sdumitriu: That's not good...

similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/DefaultPatientSimilarityScorer.java
```diff
@@ -0,0 +1,147 @@
...
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import javax.inject.Singleton;
+
+/**
+ * For Arun.
```
sdumitriu: Not good.

similarity-data-api/src/main/java/org/phenotips/data/similarity/PatientSimilarityScorer.java
```diff
@@ -0,0 +1,41 @@
...
+ * Phenotype similarity scorer.
+ *
+ * @version $Id$
+ * @since 1.4M1
+ */
+@Role
+public interface PatientSimilarityScorer
+{
+    /**
+     * Returns phenotype similarity score.
```
sdumitriu: The name of the class suggests that this is a generic similarity scorer, not one that only looks at phenotypes. Either:

- keep the API to return just a single score, but then modify the javadoc to be more generic, and write different implementations for different score aspects
- modify the API to return more than one score, possibly as a map (`{"phenotype":0.6, "genotype": 1.0, "disease":0,"overall":0.8}`)

PR #262, opened 2019-04-11
PN-457: My Matches page is not showing matching genes if one side has vcf file uploaded

veronikaslc:
Fixed.

### DIFF
--- a/matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
+++ b/matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
@@ -55,6 +55,7 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
 import org.json.JSONArray;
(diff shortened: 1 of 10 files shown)


BrianLiSK:
Looks good to me in terms of fixing the bug. The emails seem okay too. Anything else I brought up is probably outside the scope for this fix.

matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
```diff
@@ -149,26 +154,27 @@
      * @param match the match that contains the patient this object represents
      * @param patient the patient that this object represents
      * @param serverId id of server where patient is found
+     * @param matchedGenes all matched genes between two patients
      */
-    public DefaultPatientInMatch(PatientMatch match, Patient patient, String serverId)
+    public DefaultPatientInMatch(PatientMatch match, Patient patient, String serverId, Set<String> matchedGenes)
```
sdumitriu: This seems fishy. Why do genes need to be treated in a special way?

matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
```diff
@@ -401,6 +411,20 @@ private void readDetails(Patient patient)
         }
     }
 
+    private Set<String> getMatchedExomeGenes(Set<String> matchedGenes)
+    {
+        if (this.hasExomeData()) {
+            // we assume that matched exome genes are all that are not candidate/solved
```
sdumitriu: Assumptions are bad. Can't this info be made explicit? By using a specific status?

PR #265, opened 2019-04-24
PN-466: Visited UCSC Genome Browser links are displayed in white fonts

veronikaslc:
Fixed

### DIFF
--- a/ui/src/main/resources/PhenoTips/SimilarCases.xml
+++ b/ui/src/main/resources/PhenoTips/SimilarCases.xml
@@ -1659,6 +1659,9 @@
     color: $theme.notificationErrorColor;
     font-size: .87em;
 }
+.match-details .variants-data a:visited {
+    color: #000000 !important;
+}
 .match-details table.phenotype-matches .query,
 .match-details table.phenotype-matches .result {
   text-align: left;


ui/src/main/resources/PhenoTips/SimilarCases.xml
```diff
@@ -1659,6 +1659,9 @@
     color: $theme.notificationErrorColor;
     font-size: .87em;
 }
+.match-details .variants-data a:visited {
+    color: #000000 !important;
```
BrianLiSK: Isn't this black? Are we sure we don't want dark grey for visited ones?

marta-: The UCSC links are implemented as "secondary action" buttons. You should use the color defined in the color theme.
```suggestion
    color: $theme.buttonSecondaryTextColor !important;
```

ui/src/main/resources/PhenoTips/SimilarCases.xml
```diff
@@ -1659,6 +1659,9 @@
     color: $theme.notificationErrorColor;
     font-size: .87em;
 }
+.match-details .variants-data a:visited {
```
marta-: This is not the right place to fix the issue. It is actually caused by a selector that is too broad in some unrelated code: https://github.com/phenotips/phenotips/blob/master/components/gene-panels/ui/src/main/resources/PhenoTips/PanelsDisplay.xml#L552 . That selector can affect much more than just the UCSC links for variants. The problem should be fixed there directly.

marta-:
Closing this PR since the root cause is in [PhenoTips code ](https://github.com/phenotips/phenotips/blob/master/components/gene-panels/ui/src/main/resources/PhenoTips/PanelsDisplay.xml#L552) and that's where it should be fixed.

PR #270, opened 2019-05-01
PN-446: Refactor the /rest/patients/matching-notifications APIs to be RESTful

veronikaslc:


matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/internal/DefaultMatchingNotificationResource.java
```diff
@@ -262,70 +259,41 @@ public Response getEmailToBeSent(String matchId, String patientId, String server
...
-            return Response.status(Response.Status.BAD_REQUEST).build();
+            if (match != null) {
+                return Response.ok(match.toJSON(), MediaType.APPLICATION_JSON_TYPE).build();
+            } else {
+                return Response.status(Response.Status.INTERNAL_SERVER_ERROR).build();
+            }
+        } catch (final AccessControlException e) {
+            this.slf4Jlogger.error("No rights to modify match with id [{}]", matchId);
+            return Response.status(Response.Status.FORBIDDEN).build();
         }
```
allasm: Just in case i would add a generic catch(Exception) as well

matching-notification-api/src/main/java/org/phenotips/matchingnotification/internal/DefaultMatchingNotificationManager.java
```diff
@@ -205,85 +201,57 @@ public boolean saveOutgoingMatches(List<? extends PatientSimilarityView> similar
     }
 
     @Override
-    public boolean setStatus(Set<Long> matchesIds, String status)
+    public PatientMatch setStatus(Long matchId, String status)
```
allasm: Since a try-catch was removed, this method can now throw - either some unknown exception which can always happen, or (more likely) an AccessControlExfception. I would document that explicitly

matching-notification-api/src/main/java/org/phenotips/matchingnotification/internal/DefaultMatchingNotificationManager.java
```diff
@@ -205,85 +201,57 @@ public boolean saveOutgoingMatches(List<? extends PatientSimilarityView> similar
...
-                iterator.remove();
-            }
+        List<PatientMatch> matches = this.matchStorageManager.loadMatchesByIds(Collections.singleton(matchId));
+        if (matches.size() == 0) {
+            this.logger.error("No matches found for matchId " + matchId);
+            return null;
+        }
+        PatientMatch match = matches.get(0);
+        if (!currentUserHasViewAccess(match)) {
+            throw new AccessControlException("Current user has no rights to notify matchId " + matchId);
```
allasm: I would add a `logger.error()` here as well, this is unlikely to happen and it is useful to see this in the logs for debugging

matching-notification-migrations/src/main/java/org/phenotips/matchingnotification/internal/R74696PatientNetwork456DataMigration.java
```diff
@@ -25,7 +25,6 @@
 import javax.inject.Singleton;
```
allasm: Why is this file included in this PR?

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/MatchingNotificationResource.java
```diff
@@ -126,8 +129,8 @@ Response getMatches(@FormParam("minScore") @DefaultValue("0.5") double minScore,
     @POST
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
-    @Path("/send-user-notifications")
-    Response sendUserNotification(@FormParam("matchId") String matchId,
+    @Path("/{matchId}")
+    Response sendUserNotification(@PathParam("matchId") Long matchId,
```
allasm: In the issue description on JIRA I proposed to move all email-related functionality to `/matches/${matchId}/email` (both POST and GET). Lets wait until that is resolved before we merge the issue and finalize documentation

allasm: @veronikaslc changes were approved

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/internal/DefaultMatchingNotificationResource.java
```diff
@@ -405,4 +405,31 @@ private boolean isCurrentUserAdmin()
...
+    {
+        if (matchId == null) {
+            this.slf4Jlogger.error("The requested match id is blank");
+            return Response.status(Response.Status.BAD_REQUEST).build();
+        }
+
+        try {
+            PatientMatch match = this.matchingNotificationManager.getMatch(matchId);
+            JSONObject matchJson = new JSONObject();
+            matchJson.put("results", match != null ? match.toJSON() : null);
```
allasm: i wouldn't overcomplicate this and would just return `match.toJSON()` as the response JSON, with no wrappers and results fields, similar to how patient REST GET request works

PR #274, opened 2019-05-08
PN-474 (stable): When searching for similar patients all patients with matching genes should be included in the matched list

allasm:
Before this PR can be merged it should be rebased on top of PC which is based on a PT with PT-4004 included. Also some debug log messages useful for testing this PR should be removed before merging. Also need a copy of this PR for `master`


allasm:
@sashaandjic This was due to default row limit on SOLR queries. Fixed

PR #276, opened 2019-05-14
PN-231: Move all patient matches to a separate tab on the patient page

veronikaslc:
Done

### DIFF
--- a/ui/src/main/resources/PhenoTips/SimilarCases.xml
+++ b/ui/src/main/resources/PhenoTips/SimilarCases.xml
@@ -1254,7 +1254,7 @@
       });
     } else {
       // Matched patiens tab is not active
-      // generate the tab for similar cases only if the "real" conscent is granted
+      // generate the tab for similar cases only if the "real" consent is granted
 
       // if consents are already available from the patient page
       if ($('patient-consents-json')) {
--- a/ui/src/main/resources/PhenoTips/SimilarCases.xml
+++ b/ui/src/main/resources/PhenoTips/SimilarCases.xml
@@ -57,7 +57,6 @@
       #set ($min_score = "$!{suggestionsConfig.getProperty('min_score').value}")
       ##
       (% class="clear clinical-info similarity-info chapter"%)(((##
-      == $escapetool.xml($services.localization.render('phenotips.similarCases.title')) ==
       {{html clean=false}}&lt;div class="hidden"&gt;&lt;input type="hidden" id="similarity_results__max_results" value="${max_results}"/&gt;&lt;input type="hidden" id="similarity_results__min_score" value="${min_score}"/&gt;&lt;/div&gt;{{/html}}##
       (% class="similar-cases-category similar-cases-local" %)(((
       === $escapetool.xml($services.localization.render('phenotips.similarCases.subtitle')) ===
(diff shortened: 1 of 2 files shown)


ui/src/main/resources/PhenoTips/SimilarCases.xml
```diff
@@ -1240,7 +1336,7 @@
       <parse>0</parse>
     </property>
     <property>
-      <use>onDemand</use>
+      <use>always</use>
```
allasm: use `always` seems wrong (things may not work with "onDemand" or "currentPage", but maybe we need to refactor it some other way instead?)

sashaandjic:
It looks very good. The only thing that may improve the look is maybe to have the colour of inactive tab orange (all links and buttons, clickable thing, are orange in PC) and maybe use "Matches" title instead of "Simmilar cases" so it is consistent with the My Matches.
Does the refresh matches button takes care of all local and remote matches?

image

veronikaslc:
@sashaandjic inactive tabs are gray by the app schema, should not be orange. I would not use "Matches" title instead because the "Similar cases" tab shows only cases similar to the opened patient. "My Matches" page shows **all** matches for **all** patients. They are different things.

ui/src/main/resources/PhenoTips/SimilarCases.xml
```diff
@@ -1090,23 +1144,93 @@
...
+      if (requiredNotGranted.length &gt; 0) {
+        return;
+      }
+    }
+
+    var _generateSimilarCasesTab = function (isActive) {
+      return _generateTab(
+        'similar-cases-tab', // element id
+        isActive,
+        'Similar Cases', //displayed text
```
marta-: Would be nice if this was translatable. You could use the[ `phenotips.similarCases.title`](https://github.com/phenotips/patient-network/blob/39217da5d46ea45c6917c0052ecdee9e30e52b7e/similarity-data-impl/src/main/resources/ApplicationResources.properties#L1) key that was used for the heading I suggested to remove on line 60. I would also [change the translation for this key](https://github.com/phenotips/patient-network/blob/39217da5d46ea45c6917c0052ecdee9e30e52b7e/similarity-data-impl/src/main/resources/ApplicationResources.properties#L1) to be "Matches" or "Matches for this patient" or "Matches for <patient id> (to be discussed with others) so that it's more consistent with the terminology we use on the My Matches page.

marta-:
> The only thing that may improve the look is maybe to have the colour of inactive tab orange (all links and buttons, clickable things, are orange in PC)

I was actually thinking the same thing when I looked at the VM. the tab needs to pop more, or else the user will completely miss it. I don't think it's a bad idea to at least try to make it the same color as the "My Matches" button from the homepage widget and see how that works.

marta-:
@veronikaslc Please see the UI suggestions below:

"Matching patient profile" tab orange and with updated title when inactive:
image

When "Matching patient profile" is active:
* the usual patient record header (authors) and menu are not shown, since they're confusing on a page with no actual patient data
* the inactive tab which allows to return to the patient record is light gray
* the "Refresh matches" button for local matches is aligned with the "Matches on PC" title
image

@veronikaslc @sashaandjic Thoughts?

sashaandjic:
I would also make inactive Patient tab (second screenshot) orange.

sashaandjic:
"Matching patients" instead of "Matching patients' profiles" ?

marta-:
> I would also make inactive Patient tab (second screenshot) orange.

Ok, I have no strong opinion one way or the other,

marta-:
> "Matching patients" instead of "Matching patients' profiles" ?

It's shorter, so why not

sashaandjic:
> > I would also make inactive Patient tab (second screenshot) orange.
> 
> Ok, I have no strong opinion one way or the other,

I dislike gray for some reason. Also, we can stick to orange/black tabs for consistency (orange is clickable, gray disabled).

veronikaslc:
image

veronikaslc:
can test on this VM http://172.20.4.111:8080

ui/src/main/resources/PhenoTips/SimilarCases.xml
```diff
@@ -1684,137 +1807,108 @@
...
     </property>
     <property>
-      <extensionPointId>org.phenotips.patientSheet.after</extensionPointId>
+      <code>#template("colorThemeInit.vm")
+
+#document-head .title-tab:not(.active) {
+  background-color: $theme.buttonPrimaryBackgroundColor !important;
+  background-image: linear-gradient(to bottom,$theme.buttonPrimaryGradientColor 0,$theme.buttonPrimaryBackgroundColor 100%) !important;
+  font-weight: normal !important;
+  margin: 0 0 0 1px !important;
```
marta-: There should be a margin right applied to all tabs instead of a margin left applied to non-active ones.
```suggestion
}
#document-head .title-tab {
  margin: 0 1px 0 0 !important;
```

veronikaslc:
ready to test at http://172.20.4.126:8080/

sashaandjic:
Looks great. For some reason sometimes when I click on the Matching Patients tab I get this:
image

I doesn't matter if I was observing an existing patient or creating a new one. I cannot seem to find the common thread. Most often it happens once right after loggin in and then clicking on a patient from Browse all patient grid. Mind boggling...

sashaandjic:
Should we make the server names more prominent? Maybe add a line between them? Or have them in a large grid? Here is the look with only one MME server:
image

ui/src/main/resources/PhenoTips/SimilarCases.xml
```diff
@@ -1684,137 +1817,126 @@
...
-      <extensionPointId>org.phenotips.patientSheet.after</extensionPointId>
+      <code>#template("colorThemeInit.vm")
+
+.matching-patients #mainContentArea {
+  border-top: .4em solid $theme.menuGradientColor !important;
+}
+.matching-patients #document-head {
+  margin-top: -3.3em !important;
+}
+.space-data #document-head .title-tab.active, .space-data #document-head .title-tab:not(.active) {
```
marta-: why not just `.space-data #document-head .title-tab`?
```suggestion
.space-data #document-head .title-tab {
```

marta-:
> Should we make the server names more prominent?

@sashaandjic You are right, however I wouldn't put too much effort into it at the moment, since the next step is to replace this UI with the one from "My matches" table and stop maintaining two different UIs that are supposed to do the same thing. It's true that if the server names are **bolded** it looks better, so if @veronikaslc wants to do that then great, otherwise we can also leave it as it is for now.

veronikaslc:
@sashaandjic please, retest

PR #277, opened 2019-05-14
PN-475: Inconsistent rules when displaying externalIDs of the matched record on Patient form and on My Matches page

veronikaslc:
Fixed

### DIFF
--- a/matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
+++ b/matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
@@ -85,6 +85,8 @@ public class DefaultPatientInMatch implements PatientInMatch
 
     private static final GroupManager GROUP_MANAGER;
 
+    private static final AccessLevel VIEW;
+
     private static final String GENES = "genes";
 
     private static final String MATCHED_EXOME_GENES = "matchedExomeGenes";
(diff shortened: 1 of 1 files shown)


matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
```diff
@@ -26,9 +26,11 @@
 import org.phenotips.data.permissions.AccessLevel;
 import org.phenotips.data.permissions.EntityPermissionsManager;
 import org.phenotips.data.permissions.internal.EntityAccessManager;
+import org.phenotips.data.similarity.AccessType;
 import org.phenotips.data.similarity.PatientGenotype;
 import org.phenotips.data.similarity.PatientGenotypeManager;
 import org.phenotips.data.similarity.PatientSimilarityView;
+import org.phenotips.data.similarity.internal.DefaultAccessType;
```
allasm: We can check view access to a `Patient` without using classes from other internal package - not sure why similarity uses those (I think because there it deals with two patients, - but we only have on here). Here we have standard access managers available which have nothing to do with similarity and which can tell if current user has view access. In fact I think `this.access` already has all you need directly, and use it e.g. on line 221 below

veronikaslc: If we are aiming patient match page to be consistent with similar matches, we need to use the same methods to determine the access to the patient external ID. RestrictedSimilarityView uses DefaultAccessType for that purpose. If we here will use different method it is prone to future bugs in case DefaultAccessType will change and we will have 2 different methods.

allasm:
We are completely removing all similarity view related classes soon. So we should reproduce the functionality here, but usi g new better code. Then later similarity views will be removed and this code will be used instead, so there will only be one code path, this one and not that one. So it is better to make it correct and simple and not rely on code that will be removed

veronikaslc:
We shell not rely now on what we may or may no do in future. The correct approach is make changes when they belong. When we will remove similarity view related classes, we should will simplify related code then and not now.

veronikaslc:
@allasm comments addressed, force-pushed changes

PR #278, opened 2019-05-30
PN-278: When sending notifications for both incoming and outgoing versions of a match, an error is reported when it should not be

veronikaslc:
Fixed

### DIFF
--- a/matching-notification-api/src/main/java/org/phenotips/matchingnotification/internal/DefaultMatchingNotificationManager.java
+++ b/matching-notification-api/src/main/java/org/phenotips/matchingnotification/internal/DefaultMatchingNotificationManager.java
@@ -33,6 +33,7 @@ import org.xwiki.component.annotation.Component;
 import java.security.AccessControlException;
 import java.sql.Timestamp;
 import java.text.SimpleDateFormat;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
@@ -86,14 +87,17 @@ public class DefaultMatchingNotificationManager implements MatchingNotificationM
 
         for (PatientMatchEmail email : emails) {
 
-            List<PatientMatchNotificationResponse> notificationResults = this.notifier.notify(email);
+            PatientMatchNotificationResponse notificationResult = this.notifier.notify(email);
 
-            List<PatientMatch> successfulMatches = this.getSuccessfulNotifications(notificationResults);
+            if (notificationResult.isSuccessul()) {
+                Collection<PatientMatch> updatedMatches =
+                    this.updateNotificationHistory(notificationResult, email, "notification");
+                notificationResult.setPatientMatches(updatedMatches);
+            }
 
-            this.updateNotificationHistory(successfulMatches, email, "notification");
-
-            responses.addAll(notificationResults);
+            responses.add(notificationResult);
         }
+
         return responses;
     }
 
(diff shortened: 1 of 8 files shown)


sashaandjic:
on http://172.20.4.126:8080 when sending notifications to the matches P0000002-P0000026(MME) and P0000026(MME)-P0000002
image
I get this
image

veronikaslc:
PR not ready yet for testing, in progress.

veronikaslc:
@sashaandjic ready to test

sashaandjic:
I still get an error msg :
image

sashaandjic:
I logged as John and placed "32" in Patient ID contains. Here is what I got (it appears as thw same match was duplicated):
image

Then I chose the bottom one to and clicked on CONTACT button. The communication history shows 2 emails being sent:
image

veronikaslc:
@sashaandjic Fixed, ready for re-test

sashaandjic:
The defect is fixed. Now I can select both incoming and outgoing match and notify them with no error.

However, I'm not sure if the following is intended behaviour. I wanted to notify only one side of the match. I selected one email address and hit the SEND NOTIFICATIONS button.
image
The result: both sides got notification history. The one that I picked has reference email (one cell) shaded in darker colour. The one that I was not going to notify is shaded in lighter colour (the whole row).
image

This behaviour in a way makes sense. I just need to verify this with the rest of the team before I can approve it.
@allasm @marta-

matching-notification-api/src/main/java/org/phenotips/matchingnotification/internal/DefaultMatchingNotificationManager.java
```diff
@@ -86,14 +87,17 @@
...
-            List<PatientMatchNotificationResponse> notificationResults = this.notifier.notify(email);
+            PatientMatchNotificationResponse notificationResult = this.notifier.notify(email);
 
-            List<PatientMatch> successfulMatches = this.getSuccessfulNotifications(notificationResults);
-
-            this.updateNotificationHistory(successfulMatches, email, "notification");
+            if (notificationResult.isSuccessul()) {
+                Collection<PatientMatch> updatedMatches =
+                    this.updateNotificationHistory(notificationResult, email, "notification");
+                notificationResult.setPatientMatches(updatedMatches);
```
allasm: this whole design where we pass `notificationResult` into a method, get something back, which is then only used to update `notificationResult` outside the method looks iffy, updating it inside `updateNotificationHistory()` seems more proper. Not critical though

matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
```diff
@@ -563,21 +592,41 @@ public boolean setUserContacted(PatientMatch match, boolean isUserContacted)
...
+    public Collection<PatientMatch> updateNotificationHistory(PatientMatch match, JSONObject notificationRecord)
     {
         Session session = this.beginTransaction();
         boolean transactionCompleted = false;
 
+        Collection<PatientMatch> matchesToUpdate = new ArrayList<>();
+        matchesToUpdate.add(match);
+
+        if (!match.isLocal()) {
+            String referenceServerId = match.isIncoming() ? match.getReferenceServerId() : "";
```
allasm: use `getStoredServerId()` instead of explicit check for incoming\outgoing:
`String referenceServerId = this.getStoredServerId(match.getReferenceServerId())`

matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
```diff
@@ -502,6 +503,34 @@ private void copyToHistory(Session session, Collection<PatientMatch> matches)
...
+    private PatientMatch loadMatchBetweenPatients(String referencePatientId, String referenceServerId,
+        String matchedPatientId, String matchedServerId)
+    {
+        if (StringUtils.isNotEmpty(referencePatientId)) {
+            // referencePatientId is "reference patient"
+            Criterion directMatch = Restrictions.and(
+                this.patientIsReference(referencePatientId, referenceServerId),
+                this.patientIsMatch(matchedPatientId, matchedServerId));
+            List<PatientMatch> matches = this.loadMatchesByCriteria(directMatch);
+            if (matches.isEmpty()) {
```
allasm: it is better (less `returns` and no `else`) to
```
if (!matches.isEmpty()) {
  return matches.get(0);
}
```
then `return null;` outside all `if`-s

matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
```diff
@@ -502,6 +503,34 @@ private void copyToHistory(Session session, Collection<PatientMatch> matches)
...
+    /**
+     * Load {@code match} with specified parameters.
+     *
+     * @param referencePatientId id of reference patient
+     * @param referenceServerId id of the server that hosts reference patient
+     * @param matchedPatientId id of matched patient
+     * @param matchedServerId id of the server that hosts matched patient
+     * @return match
+     */
+    private PatientMatch loadMatchBetweenPatients(String referencePatientId, String referenceServerId,
```
allasm: maybe place it next to `loadMatchesbyId()` in the file (since the methods are related) or next to `updateNotificationHistory()` since that is the only method that uses it? Otherwise it is in a random place in the file

matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/internal/DefaultMatchStorageManager.java
```diff
@@ -563,21 +592,41 @@ public boolean setUserContacted(PatientMatch match, boolean isUserContacted)
...
+
+        Collection<PatientMatch> updatedMatches = new ArrayList<>();
         try {
-            match.updateNotificationHistory(notificationRecord);
-            session.update(match);
-            transactionCompleted = true;
+            for (PatientMatch matchToUpdate : matchesToUpdate) {
+                matchToUpdate.updateNotificationHistory(notificationRecord);
+                session.update(matchToUpdate);
+                transactionCompleted = true;
```
allasm: `transactionCompleted = true;` should only be used if/when all matches were updated, not just one of many in the loop. If any one failed, we should roll back the entire transaction

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/internal/DefaultMatchingNotificationResource.java
```diff
@@ -218,7 +220,10 @@ public Response sendUserNotification(Long matchId, String patientId, String serv
             if (notificationResult.isSuccessul()) {
                 result = this.successfulIdsToJSON(Collections.singletonList(matchId),
                     Collections.singletonList(matchId));
-                PatientMatch match = notificationResult.getPatientMatch();
+
+                Collection<PatientMatch> matches = notificationResult.getPatientMatches();
+                PatientMatch match = matches.iterator().next();
```
allasm: lets talk about the logic of taking the first match only here

PR #279, opened 2019-05-30
PN-483 (stable): The migrator for match comments saves the comments in invalid json format

veronikaslc:
Fixed

### DIFF
--- a/matching-notification-migrations/src/main/java/org/phenotips/matchingnotification/internal/R74693PatientNetwork377DataMigration.java
+++ b/matching-notification-migrations/src/main/java/org/phenotips/matchingnotification/internal/R74693PatientNetwork377DataMigration.java
@@ -18,20 +18,21 @@
 
 package org.phenotips.matchingnotification.internal;
 
-import org.phenotips.matchingnotification.match.PatientMatch;
-
 import org.xwiki.component.annotation.Component;
-import org.xwiki.model.reference.DocumentReferenceResolver;
-import org.xwiki.model.reference.EntityReferenceSerializer;
+
+import java.math.BigInteger;
+import java.util.List;
 
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
 
 import org.hibernate.HibernateException;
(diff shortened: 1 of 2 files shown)


matching-notification-migrations/src/main/java/org/phenotips/matchingnotification/internal/R74693PatientNetwork377DataMigration.java
```diff
@@ -90,7 +90,7 @@ public void hibernateMigrate() throws DataMigrationException, XWikiException
 
         try {
             Query q1 = session.createQuery("update " + PatientMatch.class.getCanonicalName()
-                + " set comments = concat('[{ comment:', comment, '}]') where comment <> ''");
+                + " set comments = concat('[{ \"comment\":\"', comment, '\"}]') where comment <> ''");
```
marta-: The comment itself needs to be escaped before putting it in JSON. Quotes in the text can make the JSON invalid.

matching-notification-migrations/src/main/java/org/phenotips/matchingnotification/internal/R74693PatientNetwork377DataMigration.java
```diff
@@ -89,9 +81,19 @@ public void hibernateMigrate() throws DataMigrationException, XWikiException
         Transaction t = session.beginTransaction();
 
         try {
-            Query q1 = session.createQuery("update " + PatientMatch.class.getCanonicalName()
-                + " set comments = concat('[{ comment:', comment, '}]') where comment <> ''");
-            q1.executeUpdate();
+            Query q = session.createQuery("from " + PatientMatch.class.getCanonicalName() + " where comment <> ''");
```
allasm: This query will break when `comment` field is un-mapped from the PatientMatch class. Also this query will likely break in `1.3` since there will be multiple implementations of the `PatientMatch` interface and not sure which table the query will use then (need to specify that explicitly here I think). We discussed this with Veronika already, and will make changes

PR #281, opened 2019-06-03
PN-484: Unexpected duplicate matches in matching notification database

allasm:
Fixed, reverted behaviour to PC 1.1.x level (a better fix is implemented as part of PN-401)


sashaandjic:
Not sure if it related to this PR, but when I try to add to a discussion I get this:
image
If I ignore it, the comment will be saved

PR #287, opened 2019-07-02
PN-490: Use matching notification widget on similar cases patient tab

veronikaslc:
Done.


sashaandjic:
I don't like the look and feel of the new implementation. Users will definitely prefer the old layout. Here are some suggestions, however I since they are quite extensive I do not believe they are feasible:
1. Both Reference columns are unnecessary and should be removed
2. the first 4 filter fields are the least important in this context; they should either be removed or moved to the bottom
3. Server location selector should go to the top right below the Minimum match scores
4. Server location selector and REFRESH MATCHES should be merged together (e.g. refresh buttons need to be shown next to the server names)
5. in the Matched column I would remove the last two email addresses as they are exactly the same for every match (they should be included in the mail though)
6 adding Server column and enabling sorting would be highly beneficial

PN-490

marta-:
All those changes are nice to have, but we should focus on  what's really important.

I would:
* Hide the contact reference column **on the patient page**
* Hide the "Hide own solved cases" **on the patient page**
* Move MME filters to the top **for everyone, both in "My matches" and on patient page**

sashaandjic:
The current look needs some improvements:
- Match location title should be on the top boxed (orange box)
- MATCH SEARCH CRITERIA box should be below MATCH LOCATION
- Remove from patient page: patient id contains, gene symbol contains, phenotype and email contains, VCF data uploaded filters
image

@marta fdyt?

sashaandjic:
patient form

sashaandjic:
Here is the look of the current implementation vs proposed. @marta- please see if you are happy with the current one.
current vs proposed

allasm:
@sashaandjic I like what we currently have in the VM better than what you propose.

On the "my matches" page (not patient page) the score filter should be the top section as it is right now, as its effect overwrites all other filters, and it is important to highlight that IMHO. We also had that page that way for  awhile now, and no one really complained - so I agree moving the server list to the top of the filters is good, but otherwise I would keep the page "as is".

For the patient page, most of your screenshots are irrelevant since they include lots of fields not shown on the patient page

One thing I like is separating the serve list into its own section with an orange header (the "match location" header on your screenshots)

sashaandjic:
Can we do the MATCH LOCATION section with orange header? Currently it is located under FILTERS header.

sashaandjic:
The final ui modifications are logged in a separate task PN-497.

sashaandjic:
Quite often I get "The requested page could not be found" msg when I click on the Matching patient tab right after opening the patient page. Can we make this tab clickable when the actual page is ready?
image

veronikaslc:
@sashaandjic both fixed, same VM to re-test 172.20.4.77

matching-notification-api/src/main/java/org/phenotips/matchingnotification/finder/internal/AbstractMatchFinder.java
```diff
@@ -100,9 +107,7 @@
 
     private DateTimeFormatter dateFormatter = ISODateTimeFormat.dateTime().withZone(DateTimeZone.UTC);
 
-    protected abstract Set<String> getSupportedServerIdList();
-
-    protected abstract MatchRunStatus specificFindMatches(Patient patient, String serverId,
+    protected abstract Response specificFindMatches(Patient patient, String serverId,
```
allasm: Having matching code to return an object related to web (`Response`) seems wrong. I don't have a nice a quick solution how to do this otherwise, but it adds an unnecessary dependency on `javax.ws.rs.core` to a piece of code which should not care about that. A not-perfect other solution is to have our own "status" object, with similar functionality to a `Response`, and do a conversion from that object to a Response in REST code. @marta, what do you think, which design patter is better (one of these two, or something else)?

matching-notification-api/src/main/java/org/phenotips/matchingnotification/finder/internal/DefaultMatchFinderManager.java
```diff
@@ -70,6 +72,22 @@ public void findMatchesForAllPatients(Set<String> serverIds, boolean onlyCheckPa
...
     }
 
+    @Override
+    public Response findMatchesForPatient(Patient patient, String serverId)
+    {
+        this.logger.error("Finding matches for patient [{}] for server [{}]", patient.getId(), serverId);
+
+        for (MatchFinder matchFinder : this.matchFinderProvider.get()) {
+            Set<String> supportedServers = matchFinder.getSupportedServerIdList();
+            if (!supportedServers.contains(serverId)) {
```
allasm: This part (checking if server is supported) is done inside `matchFinder.findMatches()` for the other `findMatches()` call, for consistency it makes more sense to do it there for this version as well. Also i would avoid overloading the method and would call it `findmatchesForASinglePatient()` or something along those lines

matching-notification-api/src/main/java/org/phenotips/matchingnotification/storage/MatchStorageManager.java
```diff
@@ -49,7 +50,8 @@
      *        if {@code null}, then no upper limit on the match date is considered
      * @return a list of matches
      */
-    List<PatientMatch> loadMatches(double score, double phenScore, double genScore, boolean onlyCurrentUserAccessible,
+    List<PatientMatch> loadMatches(String patientId, double score, double phenScore, double genScore,
```
allasm: I would add a separate call for loading matches for a given patient, with no score etc. etc. parameters, since we should always load all available matches for a patient. Internally they both can call the same private method with different parameters, if it makes life easier, but in API I would separate the two

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -19,6 +19,13 @@ var PhenoTips = (function (PhenoTips) {
     {
         this._tableElement = $('matchesTable');
         this._ajaxURL = XWiki.contextPath + "/rest/matches";
+
+        this._loadMatchesURL = this._ajaxURL;
+        this._onSimilarCasesPage = $('similar-cases-container');
```
allasm: I would rename all reference to "similarCasesPage" to simply "patientPage". This is the first time we introduce this name\word in this code, lets not use legacy names which will be completely unclear in the future, since `SimilarCases.xml` is removed in the same PR (so in the future there will be a question "what the hell is 'similar cases'"?

allasm: Also IMHO the check for that page can be more robust than checking `$('similar-cases-container')` - e.g. check that space is data, which is less likely to change

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -120,6 +129,8 @@ var PhenoTips = (function (PhenoTips) {
         $$('th[data-column="foundTimestamp"]')[0] && $$('th[data-column="foundTimestamp"]')[0].on('click', function(event) {this._sortByColumn('foundTimestamp');}.bind(this));
 
         Event.observe(window, 'resize', this._buildTable.bind(this));
+
+        Event.observe(document, "matches:refreshed", this._showMatches.bind(this));
```
allasm: just curious what is different on patient page that we need this even handler now, but did not need it in the past?

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -1336,13 +1384,12 @@ var PhenoTips = (function (PhenoTips) {
 
     _getPatientToBeContactedByCurrentNonAdminUser : function(match)
     {
-        if (!this._accessAtLeastEdit(match.reference) && !this._accessAtLeastEdit(match.matched)) {
-            // user does not have at least edit access to any of the patients in the match: do not allow to contact or mark as user-contacted
+        if (this._isAdmin) {
```
allasm: what is the logic behind the changes in this method?

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/MatchingNotificationResource.java
```diff
@@ -93,6 +110,31 @@ Response getMatches(@FormParam("minScore") @DefaultValue("0.5") double minScore,
...
+     *        will be returned; if {@code null} or an empty string, then no lower limit on the match date is considered
+     * @param toDate if passed a date in the {@code yyyy-MM-dd} format, then only matches found on or before this date
+     *        will be returned; if {@code null} or an empty string, then no upper limit on the match date is considered
+     * @return a response containing a JSON object with a list of matches
+     */
+    @GET
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/patients/{patientId}")
+    Response getMatchesForPatient(@PathParam("patientId") String patientId,
```
allasm: I think we decided that for a single patient we don't need all those parameters, we just load all available matches for the patient. I can see an argument for keeping `fromDate` and `toDate` (when used via scripts to load only new matches), but for a single patient doing extra filtering on the back-end doesn't make much sense I think. We also talked about removing the score fields form the UI, so it will be impossible to change it as well. @marta- ?

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
veronikaslc: SimilarCases.xml is not removed

ui/src/main/resources/PhenoTips/SimilarCases.xml
```diff
@@ -39,56 +39,53 @@
...
+#set ($preferenceDoc = $xwiki.getDocumentAsAuthor("XWiki.XWikiPreferences"))
+#set ($allRemotes = $preferenceDoc.getObjects("PhenomeCentral.RemoteMatchingServiceConfiguration"))
+#foreach ($config in $allRemotes)
+  #if ($config.getProperty('remoteAuthToken').getValue() != "" &amp;&amp; $config.getProperty('serverId').getValue() != "" &amp;&amp; $config.getProperty('includeInUI').getValue() == 1)
+    #set ($serverID = $config.getProperty('serverId').getValue())
+    #set ($serverName = $config.getProperty('humanReadableName').getValue())
+    #if ($serverName == "")
+      #set ($serverName = $serverID)
+    #end
+    &lt;a class="${escapetool.xml($serverID)}" href="#"&gt;${escapetool.xml($serverName)}&lt;/a&gt;
```
allasm: Not sure which line to comment, but IMHO the dropdown for refreshing matches is a bit unintuitive, with the arrow and text that we have now

marta-: I don't like the way it looks either. It was supposed to mimic the [Bootstrap dropdown buttons](https://getbootstrap.com/docs/4.0/components/dropdowns/) which people are used to, but the styling is too different. I would:
* increase the padding to .3em
* stick the two components (label and arrow) together
* get rid of the icon
* make sure the dropdown actually looks attaced to the button
* align the dropdown text to the right, and make sure it aligns with the button text

image

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
veronikaslc: code optimisation

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/MatchingNotificationResource.java
marta-: IMO It's OK to have all parameters for when patient id is present, in case we want to do some sort of filtering via API later. I don't see a problem with keeping them all.

matching-notification-api/src/main/java/org/phenotips/matchingnotification/finder/internal/DefaultMatchFinderManager.java
marta-: Minor, code won't be that much cleaner if the check is moved inside.

matching-notification-api/src/main/java/org/phenotips/matchingnotification/finder/internal/AbstractMatchFinder.java
marta-: The design problem is very deep, I suggest the refactoring be done separately from this PR.

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/MatchingNotificationResource.java
```diff
@@ -93,6 +110,31 @@ Response getMatches(@FormParam("minScore") @DefaultValue("0.5") double minScore,
...
+        @FormParam("minScore") @DefaultValue("0.5") double minScore,
```
allasm: default values should be 0.1 for average score, and 0 for gen/phen scores. There is no UI to lower the scores now, so we should load all matches except completely crappy ones (to avoid having 10s of pages of matches no on eis interested in)

sashaandjic:
Patient page is filtering out all matches below avg score 0.4 and it should not.
@allasm suggested changing it to 0.1.

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/MatchingNotificationResource.java
allasm: note that we wont have "pages and pages" of low-scoring matches right now, but we potentially might when we enable e.g. pubcasefinder or some other matches will return bad matches, so having at least a 0.1 limit is useful

veronikaslc: @marta

marta: @veronikaslc you got the wrong person

veronikaslc:
@marta- [this is the code](https://github.com/phenotips/patient-network/blame/master/matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/internal/DefaultMatchingNotificationResource.java#L149) place where we filter by score 0.3 for non admin users

matching-notification-rest/src/main/java/org/phenotips/matchingnotification/rest/MatchingNotificationResource.java
veronikaslc: @marta-

sashaandjic:
@marta ^

allasm:
As discussed with @marta- , the values for the scores will be as follows:

- in the front end (UI/javascript) the default for now will be 0.3 for avg.score and 0 for other scores. This way users will be able to see the same matches as they can in the all-matches page, and won't be confused about lost/new matches.

- on the back-end both the hard limit and the default REST parameters will be 0.1 for avg.score and 0 for all other scores. This way we exclude potential garbage matches and will never spam the front-end with 100s of garbage matches, yet if we ever change the scrong algorithm so that lower scores make more sense we will only have to change the front-end while back-end will pass those matches /scores through

PR #288, opened 2019-07-02
PN-486: Add Across MME checkbox to matches table filters

veronikaslc:
Done


sashaandjic:
Also, when Across MME is unchecked and checked again, the numbers of matches shown next to the remote server is out of sync. In the attached screenshot both remote servers are checked and show 0 matches.
image

PR #289, opened 2019-07-04
PN-485: Make discussion window more user friendly

veronikaslc:
Done

### DIFF
--- a/matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.css
+++ b/matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.css
@@ -475,6 +475,15 @@ td[name="contact"] {
   white-space: normal;
   line-height: 12px;
 }
+.xTooltip {
+  white-space: pre-wrap;
+  white-space: -moz-pre-wrap;
+  white-space: -pre-wrap;
+  white-space: -o-pre-wrap;
+  word-wrap: break-word;
+  font-weight: normal;
+  text-transform: initial;
+}
 .notes-container h4 {
   margin: 0;
 }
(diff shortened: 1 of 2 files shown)


matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.css
```diff
@@ -475,6 +475,15 @@ td[name="contact"] {
   white-space: normal;
   line-height: 12px;
 }
+.xTooltip {
```
sdumitriu: This seems too generic and will likely break the intended behavior for other tooltips. Can you find a more specific selector that only applies to tooltips in the matches table?

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.css
```diff
@@ -475,6 +475,15 @@ td[name="contact"] {
...
+  white-space: pre-wrap;
+  white-space: -moz-pre-wrap;
```
sdumitriu: The standard value is well supported, there is no need to use the prefixed versions.

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.css
```diff
@@ -475,6 +475,15 @@ td[name="contact"] {
...
+  white-space: -pre-wrap;
+  white-space: -o-pre-wrap;
+  word-wrap: break-word;
```
sdumitriu: You can remove this line too, since it's a workaround for IE<8 not supporting `pre-wrap`. We don't support IE7 or older.

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -1478,6 +1478,9 @@ var PhenoTips = (function (PhenoTips) {
             }.bind(this));
 
         }.bind(this));
+
+        //initiate HelpButton widgets
+        Event.fire(document, "xwiki:dom:updated", {elements : this._tableElement.select('.comment-container .xHelpButton')});
```
sdumitriu: Why don't you use the whole `_tableElement`? The code responsible for `xHelpButton` should know to look inside for it's own trigger elements.

PR #290, opened 2019-07-15
PN-413: Add information to the Reference and Matched columns to show institution/group that submitted the case

veronikaslc:
Done.

### DIFF
--- a/matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/PatientInMatch.java
+++ b/matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/PatientInMatch.java
@@ -17,6 +17,7 @@
  */
 package org.phenotips.matchingnotification.match;
 
+import org.phenotips.data.ContactInfo;
 import org.phenotips.data.Patient;
 import org.phenotips.data.permissions.AccessLevel;
 import org.phenotips.data.similarity.phenotype.PhenotypesMap;
@@ -152,4 +153,9 @@ public interface PatientInMatch
      *         or null if patient does not have genotype or genotype status is unknown.
      */
     String getGenesStatus();
+
+    /**
+     * @return patient contact information.
+     */
+    ContactInfo getContactInfo();
 }
--- a/matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
+++ b/matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
@@ -263,7 +263,7 @@ public class DefaultPatientInMatch implements PatientInMatch
         if (this.contactInfo != null) {
             return this.contactInfo.getEmails();
         }
-        return null;
+        return Collections.emptyList();
     }
 
     @Override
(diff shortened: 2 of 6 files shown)


sashaandjic:
In the attached screen shot I removed email address for John and Pat on purpose. 
Here are some suggestions that may improve the look and feel:

- change the format to display to:
  user name (email address)
  institution name

OR

- show email address when you hoover over envelope icon instead of displaying "Mark match contact to notify". The current tooltip is very confusing to the user anyway.

image

@marta wdyt?

matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/PatientInMatch.java
```diff
@@ -132,11 +133,6 @@
      */
     Patient getPatient();
 
-    /**
-     * @return patient href.
-     */
-    String getHref();
```
allasm: As discussed, for existing MME matches there may be data in the `href` column which is  not present anywhere else, so it is important not to lose it and display it as we used to display before this change

allasm: we can either write a migrator or (less work) just use the data from the column if it is available

matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
```diff
@@ -259,20 +261,18 @@ public boolean isLocal()
...
-                // MME emails may be of the form "mailto:email1,email2", need ot parse that
-                if (this.href.startsWith("mailto:")) {
-                    String emailList = this.href.replace("mailto:", "");
+                // MME emails may be of the form "mailto:email1,email2", need to parse that
+                String href = this.contactInfo.getUrl();
+                if (href.startsWith("mailto:")) {
+                    String emailList = href.replace("mailto:", "");
                     emails.addAll(Arrays.asList(emailList.split(",|;")));
                 } else {
-                    emails.add(this.href);
```
allasm: this is the line which used that piece of data (...and which used to be transferred to front-end as one of the emails)

matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
```diff
@@ -258,25 +260,10 @@ public boolean isLocal()
...
-                    emails.addAll(Arrays.asList(emailList.split(",|;")));
-                } else {
-                    emails.add(this.href);
-                }
-            }
+        if (this.contactInfo != null) {
+            return this.contactInfo.getEmails();
         }
-        return emails;
+        return null;
```
allasm: there is a change in functionality, it used to return an empty collection when there were no emails, now it returns `null`. Null often requires an extra check in the calling code, any reason not to return an empty collection here as before?

matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
```diff
@@ -498,21 +493,42 @@ private Patient getLocalPatient()
...
         }
-        PatientData<ContactInfo> data = this.patient.getData("contact");
-        if (data != null && data.size() > 0) {
-            ContactInfo contact = data.get(0);
-            if (contact != null) {
-                this.href = contact.getUrl();
+        return null;
+    }
+
+    private ContactInfo rebuildContactInfo(JSONObject contact, String oldMMEContactData)
```
allasm: if this is called via line 193 where we may have a new version of a patient (via `this.patient = getLocalPatient();`) we need to get contact info from patient as well, as it is higher priority than stored data.

I.e. the code should be something like

```
if (patient != null) {
   this.populateContactInfo()
} else if (contact != null) {
...
}
```

PR #293, opened 2019-08-07
PN-497: My Matches widget UI improvements

veronikaslc:
Done.


sashaandjic:
Found column has a mysterious arrow. When I log as admin the date is displayed in two lines.
PN-497 Admin

I am not happy with the headers. They look a bit messy. I am proposing the following look:
PN-497 Proposed look
vs current look:
PN-497 Current look

The list of proposed changes:
- Get rid of Reference, Matched, Matched<< and Contact headings
- Replace Patient details with Reference patient and Matched patient (right alligned)
- Add an s to Score
- Move Found heading up
- Replace Contact emails with Contact << ( move up << )
- Make Avg-Gen-Phen heading thinner

@Marta WDYT

marta-:
@sashaandjic Agreed, with some comments:

> Add an s to Score

Not sure this is necessary

> Make Avg-Gen-Phen heading thinner

Do you mean smaller font (e.g. same font size as the actual score values)? 

> add the same score "i" button to the Minimum match scores in the right panel that is visible to admins and group managers

I don't understand this one, sorry.

sashaandjic:
@marta-  
image

veronikaslc:
As per today's discussion with @sashaandjic changes to implement:
- set line-height for the score and date columns
- get rid of the top row in the heading
- style contact info same as in history popup
- add the same score "i" button to the "Minimum match score"
- align everything to the left and middle
- collapse emails by default

marta-:
> As per today's discussion with @sashaandjic changes to implement:

I disagree with some of these changes, please see below:

> * get rid of the top row in the heading

This leaves no explanation for the "Avg", "Gen", "Phen" columns. The better fix is to keep it as before, two heading rows, where the second row only has the "Avg", "Gen", "Phen" subheadings under the "Score" heading with the (i) button.

> * align everything to the left and middle

Just Left, no need for middle anywhere.

veronikaslc:
>  The better fix is to keep it as before, two heading rows, where the second row only has the "Avg", "Gen", "Phen" subheadings under the "Score" heading with the (i) button.

Please, specify on the screenshots the headings and their positioning per row for user and admin separately. Currently there are too many versions what should be where.

marta-:
Re-discussed in person, the following changes are to be made:

* Restore the two-level table headings:
  - Level 1: Status (for admins); Notes (added for admins, the column also includes "discussion"), **Match** (Replaces "Patient details"), Score (i), Found, Contact
  - Level 2: Reference and Matched under "Match"; Avg, Gen, Phen under "Score"; Reference and Matched under Contact; 
* All `td`s and `th`s should have vertical align: top, text-align: left. Exception: scores are center-aligned
* All `th`s should have `padding: .5em .4em` and `font-variant: small-caps`
* In the contact column: add `margin-bottom: 10px` to the affiliation line
* In the contact column for admins: replace full envelope icon with outlined envelope icon
* Ensure contact info looks consistent in the table and in the contact history popup

PR #294, opened 2019-08-13
PN-499: Disclaimers specific to all MME sites should be available on the Matches table; PN-498: In the "refresh matches" dropdown, display when (or if) the maches were last ran

veronikaslc:
Done


matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -179,6 +183,36 @@ var PhenoTips = (function (PhenoTips) {
...
+                var endTimeInput = trigger.up('label').down('.endTime');
+                var lastUpdateText = this._MATCHES_NEVER_RUN;
+                if (endTimeInput && endTimeInput.value) {
+                	var time = new Date(endTimeInput.value);
+                    lastUpdateText = this._MATCHES_LAST_RUN + " " + time.toISOString().split('T')[0];
+                }
+                disclaimerContainer.insert(new Element('div').insert(lastUpdateText));
+
+                var serverName = trigger.up('label').down('.serverName').value;
+                var disclaimerDialog = new PhenoTips.widgets.ModalPopup(disclaimerContainer, false, {'title': serverName + " $escapetool.javascript($services.localization.render('phenotips.myMatches.disclaimer.disclaimerDialogTitle'))", 'removeOnClose': false, 'resetPositionOnShow': false});
```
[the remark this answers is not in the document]
marta-: Proposed UI:

image

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -179,6 +183,36 @@ var PhenoTips = (function (PhenoTips) {
...
+                var disclaimerContainer = new Element('div');
+                disclaimerContainer.insert(new Element('p', {'class' : 'mme-disclaimer'}).insert(disclaimerTextInput.value));
+
+                var endTimeInput = trigger.up('label').down('.endTime');
+                var lastUpdateText = this._MATCHES_NEVER_RUN;
+                if (endTimeInput && endTimeInput.value) {
+                	var time = new Date(endTimeInput.value);
+                    lastUpdateText = this._MATCHES_LAST_RUN + " " + time.toISOString().split('T')[0];
+                }
+                disclaimerContainer.insert(new Element('div').insert(lastUpdateText));
```
marta-: This should be colored as specified in the Jira issue to make it stand out more.

matching-notification-api/src/main/resources/ApplicationResources.properties
```diff
@@ -181,3 +180,8 @@ phenotips.myMatches.contact.preview.error.header=Error getting contact email pre
 phenotips.myMatches.notification.dialog.success.title=Success
 phenotips.myMatches.notification.dialog.error.title=Error
 phenotips.myMatches.notification.dialog.closeButton.name=Close
+
+phenotips.myMatches.disclaimer.matchesLastRun=Matches last ran on
+phenotips.myMatches.disclaimer.noMatchRequest=Matches were never ran
```
marta-: Now that I see it in the UI I'm not sure the texts I proposed in the Jira tasks are clear enough or even correct English.

My new suggestion would be "Matches with this database were last computed on <date>." and "Matches with this database were never computed".

matching-notification-api/src/main/resources/ApplicationResources.properties
```diff
@@ -181,3 +180,9 @@ phenotips.myMatches.contact.preview.error.header=Error getting contact email pre
...
...isclaimer.disclaimer.PhenomeCentral=<p>PhenomeCentral provides data in good faith as a research tool, but without verifying the accuracy, clinical validity or utility of the data. PhenomeCentral makes no warranty, express or implied, nor assumes any legal liability or responsibility for any purpose for which the data are used.</p><br><p>The user certifies that no attempt to identify individual patients will be undertaken.</p><br><p>Prior to using PhenomeCentral data in a publication, the user will contact the owner of the matching dataset to assess the integrity of the match and if validated will offer appropriate agreed recognition of their contribution, which may include co-authorship if the magnitude of the contribution warrants it to at least one representative from the project/participating centre (possibly the member who submitted the patient data).</p><br><p>If PhenomeCentral enables a research publication, the authors must acknowledge PhenomeCentral using the following wording: &ldquo;This study makes use of data shared through the PhenomeCentral repository. Funding for PhenomeCentral was provided by Genome Canada and Canadian Institute of Health Research (CIHR).&rdquo;</p>
```
marta-: It's a bad practice to put markup in translation strings. I would have opted for a different approach:
- [ ] Place the disclaimer in its own wiki document, at `PhenomeCentral/ServiceDisclaimer`, with the title "PhenomeCentral Service Disclaimer" and following content (this is wiki markup):
```
PhenomeCentral provides data in good faith as a research tool, but without verifying the accuracy, clinical validity or utility of the data. PhenomeCentral makes no warranty, express or implied, nor assumes any legal liability or responsibility for any purpose for which the data are used.

The user certifies that no attempt to identify individual patients will be undertaken.

Prior to using PhenomeCentral data in a publication, the user will contact the owner of the matching dataset to assess the integrity of the match and if validated will offer appropriate agreed recognition of their contribution, which may include co-authorship if the magnitude of the contribution warrants it to at least one representative from the project/participating centre (possibly the member who submitted the patient data).

If PhenomeCentral enables a research publication, the authors must acknowledge PhenomeCentral using the following wording: "//This study makes use of data shared through the PhenomeCentral repository. Funding for PhenomeCentral was provided by Genome Canada and Canadian Institute of Health Research (CIHR).//"
```
- [ ] Load the content of the document like this:
```#set ($disclaimer = "$!xwiki.getDocument('PhenomeCentral.ServiceDisclaimer').getRenderedContent()")```

The advantage of having the disclaimer in its own document is that we can link it and display it as is for the users on the PC website.

File: matching-notification-ui/src/main/resources/MatchingNotification/MatchesTable.xml [hunk 5]
```diff
@@ -127,20 +129,44 @@ $xwiki.jsx.use('PhenoTips.Widgets')##
...
+                    #end
+                    #set ($disclaimer = $config.getProperty('disclaimer').getValue())
+                    #set ($matchesUpdateObject = $matchesRunDoc.getObject($infoClassname, 'serverName', $serverID))
+                    #if ($matchesUpdateObject)
+                      #set($endTime = $matchesUpdateObject.getValue('completedTime'))
+                    #end
+                    &lt;label&gt;
+                      &lt;input name="checkbox-server-id-filter" class="mme" type="checkbox" value="${escapetool.xml($serverID)}" checked&gt;
+                      &lt;input type="hidden" class="disclaimer" value="${escapetool.xml($disclaimer)}"&gt;
+                      &lt;input type="hidden" class="endTime" value="${escapetool.xml($!endTime)}"&gt;
```
marta-: Please see the screenshots below. Even though I never ran matches on the remote server called "Not matching", the a non-empty endTime is reported at it's identical to the one recorded the remote server above it, named "Local (self)".

 image

image

matching-notification-ui/src/main/resources/MatchingNotification/MatchesTable.xml
```diff
@@ -127,20 +129,44 @@ $xwiki.jsx.use('PhenoTips.Widgets')##
...
-                      &lt;div class="filter-mme-server-name"&gt;$escapetool.xml($config.getProperty('humanReadableName').getValue())&lt;span class="matches-count"&gt;&lt;/span&gt;&lt;/div&gt;&lt;br&gt;
+                    #set ($serverID = $config.getProperty('serverId').getValue())
+                    #set ($serverName = $config.getProperty('humanReadableName').getValue())
+                    #if ($serverName == "")
+                      #set ($serverName = $serverID)
+                    #end
+                    #set ($disclaimer = $config.getProperty('disclaimer').getValue())
+                    #set ($matchesUpdateObject = $matchesRunDoc.getObject($infoClassname, 'serverName', $serverID))
+                    #if ($matchesUpdateObject)
+                      #set($endTime = $matchesUpdateObject.getValue('completedTime'))
```
marta-: If the object is null, `$endTime` is not updated and retains its previous value. If matches were computed on the first remote but not on the second, the second remote will report that it had its matches computed at the same time as the first.

```suggestion
                      #set ($endTime = $matchesUpdateObject.getValue('completedTime'))
                   #else
                     #set ($endTime = '')
```

or even better, make it shorter, skip the `$matchesUpdateObject` assignment altogether and ensure the `$endTime` is always updated, with just one line:

```
#set ($endTime = "$!matchesRunDoc.getObject($infoClassname, 'serverName', $serverID).getValue('completedTime')")
```

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -91,7 +91,6 @@ var PhenoTips = (function (PhenoTips) {
         this._NOTES_TITLE = "$escapetool.xml($services.localization.render('phenotips.matchingNotifications.table.notes.title'))";
         this._NOTES_SAVE = "$escapetool.xml($services.localization.render('phenotips.matchingNotifications.table.notes.save'))";
         this._NOTES_HINT = "$escapetool.xml($services.localization.render('phenotips.matchingNotifications.table.notes.hint'))";
-        this._MATCHES_LAST_RUN = "$escapetool.xml($services.localization.render('phenotips.myMatches.disclaimer.matchesLastRun'))";
         this._MATCHES_NEVER_RUN = "$escapetool.xml($services.localization.render('phenotips.myMatches.disclaimer.noMatchRequest'))";
```
marta-: `_MATCHES_NEVER_RUN` doesn't appear to be used anywhere anymore.

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -210,11 +209,8 @@ var PhenoTips = (function (PhenoTips) {
                 disclaimerContainer.insert(new Element('div', {'class' : 'server-name'}).insert(serverName));
 
                 var endTimeInput = trigger.up('label').down('.endTime');
```
marta-: Now that the nature of the input value has changed from date to message, the variable name should change too so that the code is easier to understand. What about something like "lastMatchRunInfo"? The same thing goes for the CSS class of the input.

matching-notification-ui/src/main/resources/MatchingNotification/MatchesTable.xml
```diff
@@ -156,14 +161,17 @@ $xwiki.jsx.use('PhenoTips.Widgets')##
                       #set ($serverName = $serverID)
                     #end
                     #set ($disclaimer = $config.getProperty('disclaimer').getValue())
-                    #set ($matchesUpdateObject = $matchesRunDoc.getObject($infoClassname, 'serverName', $serverID))
-                    #if ($matchesUpdateObject)
-                      #set($endTime = $matchesUpdateObject.getValue('completedTime'))
+                    #set ($endTime = "$!matchesRunDoc.getObject($infoClassname, 'serverName', $serverID).getValue('completedTime')")
```
marta-: Now that you've moved the message generation from JS to Velocity, a different approach might be more appropriate, to avoid re-parsing the date from a string:

```
#set($endTime = '')## makes sure the previous value is erased
#set ($endTime = $!matchesRunDoc.getObject($infoClassname, 'serverName', $serverID).getValue('completedTime'))## stores the new value IF it exists
```

Then you can check if the date exists with:
```
#if ("$!endTime" != "")
````

File: matching-notification-ui/src/main/resources/MatchingNotification/MatchesTable.xml [hunk 10]
```diff
@@ -156,14 +161,17 @@ $xwiki.jsx.use('PhenoTips.Widgets')##
...
+                    #if ($endTime &amp;&amp; "$!endTime.class" == 'class java.lang.String' &amp;&amp; "$!endTime" != "")
+                      #set ($endTimeParsed = $services.jodatime.getDateTimeFormatterForPattern('EEE MMM dd kk:mm:ss z yyyy').parseDateTime($endTime).toDate())
+                      #set ($endTimeFormatted = $services.jodatime.getDateTimeFormatterForPattern($dateFormat).print($endTimeParsed))
+                      #set ($endTimeInfo = $services.localization.render('phenotips.myMatches.disclaimer.matchesLastRun', [$endTimeFormatted]))
                     #end
                     &lt;label&gt;
                       &lt;input name="checkbox-server-id-filter" class="mme" type="checkbox" value="${escapetool.xml($serverID)}" checked&gt;
                       &lt;input type="hidden" class="disclaimer" value="${escapetool.xml($disclaimer)}"&gt;
-                      &lt;input type="hidden" class="endTime" value="${escapetool.xml($!endTime)}"&gt;
+                      &lt;input type="hidden" class="endTime #if($endTime)time-updated#else never-run#end" value="${escapetool.xml($!endTimeInfo)}"&gt;
```
marta-: The `time-updated` css class is added when the matches were never computed, please see screenshot:

image

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
veronikaslc: Where do you think is better place for message generation - velocity or javascript?

marta-: I think either is fine. Generating this particular message on the backend does make it easier to handle translations properly. In general, the code is easier to understand and to maintain if the functionality is not split between frontend and backend, so you could even go one step further and generate the whole content of the tooltip on the backend before putting it in a hidden input or in a `data-` attribute. Just my 2 cents.

sashaandjic:
"Matches with this database were last..." text in disclaimer is shown in blue fonts for PhenomeCentral. For the MME servers text is shown in green. 
Disclaimers

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
veronikaslc: On the other side last run time value is used for 2 places, both in disclaimers and in dropdown for servers. Messages differ, so better shift their generation to js.

veronikaslc:
Addressed codereview comments, VM to test: 172.20.4.111

sashaandjic:
Everything looks good. Disclaimers are looking good.
There is one thing that I'm not sure how we want to handle. I added Fake PubCase Finder. I only ran matches for patient http://172.20.4.111:8080/data/P0000009 on the patient form. There are 16 matches found, however in the dropdown it says that the matches were never run. Here is the think, it is true that matches were never run for all patients, however matches were run for P0000009. What should we do? @marta 
image

sashaandjic:
Small ui issue with JohnDoe's email address 
image

sashaandjic:
Why do I see link for Mary in this case and nowhere else?
image

sashaandjic:
Matches were run on the 14th not 15th in this case
image

sashaandjic:
Should we widen the disclaimer window? I like the width we currently have in production.

sashaandjic:
Are discussion and notes icons (circled in red) a bit smaller than notification and communication history icons (circled in blue)?
image

marta-:
> Are discussion and notes icons (circled in red) a bit smaller than notification and communication history icons (circled in blue)?

Yes, they are smaller. Computed font-size 11.2px vs 14px. @veronikaslc

marta-:
> Why do I see link for Mary in this case and nowhere else?
> To reproduce login as admin. No other match has user name as link in the communication history.

I think it's because of one unmerged PR in PhenoTips.

marta-:
> Should we widen the disclaimer window? I like the width we currently have in production.

IMO what we have now on production is too wide. We could go with 50em instead of 30 if you think this is really important.

marta-:
> Small ui issues in the communication history:
> 
> * John's and Mary's email address misplaced (highlighted in red; neither had an institution at the time of contact)
> * the date is misaligned for notification and more prominently for contacts

These are unrelated to this PR, they are caused by PN-490 i think.

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -179,6 +184,63 @@ var PhenoTips = (function (PhenoTips) {
...
+                });
+                disclaimerContainer.insert(closeButton);
+
+                var serverName = trigger.up('label').down('.serverName').value;
+                disclaimerContainer.insert(new Element('div', {'class' : 'server-name'}).insert(serverName));
+
+                var endTimeInput = trigger.up('label').down('.endTime');
+                if (endTimeInput && endTimeInput.value && new Date(endTimeInput.value)) {
+                    var time = new Date(endTimeInput.value);
+                    var lastRunInfo = this._MATCHES_LAST_RUN.replace('_TIME_', time.toISOString().split('T')[0]);
```
marta-: This switches from local time zone to UTC. Matches computed at 10pm yesterday will be reported to have been computed today.

marta-:
> Matches were run on the 14th not 15th in this case

Caused by a timezone issue, see above.

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -576,7 +576,7 @@ var PhenoTips = (function (PhenoTips) {
             }.bind(this),
             onSuccess : function (response) {
                 if (response.responseJSON) {
-                    console.log("Show matches response JSON (min scores: " + options.score + "/" + options.phenScore + "/" + options.genScore + "):");
+                    console.log("Show matches response JSON (min scores: " + options.minScore + "/" + options.minPhenScore + "/" + options.minGenScore + "):");
```
marta-: There should be no `console.log()`s in production.

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -131,6 +132,8 @@ var PhenoTips = (function (PhenoTips) {
         Event.observe(window, 'resize', this._buildTable.bind(this));
 
         Event.observe(document, "matches:refreshed", this._showMatches.bind(this));
+
+        this._initiateDisclaimers();
```
marta-:
```suggestion
        this._initializeDisclaimers();
```

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -179,6 +182,54 @@ var PhenoTips = (function (PhenoTips) {
         this._currentSortingOrder = "none";
     },
 
+    _initiateDisclaimers: function() {
```
marta-:
```suggestion
    _initializeDisclaimers: function() {
```

matching-notification-ui/src/main/resources/MatchingNotification/MatchesTable.xml
```diff
@@ -127,20 +128,35 @@ $xwiki.jsx.use('PhenoTips.Widgets')##
...
             &lt;/span&gt;
             &lt;ul class="filters xform"&gt;
               &lt;li id="checkbox-server-filters"&gt;$escapetool.xml($services.localization.render('phenotips.matchingNotifications.table.serverIdCheckbox.label'))&lt;br&gt;
-                &lt;label&gt;&lt;input name="checkbox-server-id-filter" type="checkbox" value="local" checked &gt;
-                  $escapetool.xml($services.localization.render('phenotips.matchingNotifications.table.matchesSeletFilter.local'))&lt;span class="matches-count"&gt;&lt;/span&gt;&lt;br&gt;
+                &lt;label&gt;&lt;input name="checkbox-server-id-filter" type="checkbox" value="local" checked &gt;##
+                  $escapetool.xml($services.localization.render('phenotips.matchingNotifications.table.matchesSeletFilter.local'))&lt;span class="matches-count"&gt;&lt;/span&gt;##
+                  #set ($disclaimer = "$!xwiki.getDocument('PhenomeCentral.ServiceDisclaimer').getRenderedContent()")##
+                  &lt;input type="hidden" class="disclaimer" value="${escapetool.xml($disclaimer)}"&gt;##
```
marta-: The hidden inputs and the info icon only need to be generated if the disclaimer is not empty. Same comment for all remote servers.

matching-notification-ui/src/main/resources/MatchingNotification/MatchesTable.xml
```diff
@@ -127,20 +128,35 @@ $xwiki.jsx.use('PhenoTips.Widgets')##
...
+                    #set ($serverName = $config.getProperty('humanReadableName').getValue())
+                    #if ($serverName == "")
+                      #set ($serverName = $serverID)
+                    #end
+                    #set ($disclaimer = $config.getProperty('disclaimer').getValue())
+                    &lt;label&gt;
+                      &lt;input name="checkbox-server-id-filter" class="mme" type="checkbox" value="${escapetool.xml($serverID)}" checked&gt;
+                      &lt;input type="hidden" class="disclaimer" value="${escapetool.xml($disclaimer)}"&gt;
+                      &lt;input type="hidden" class="serverName" value="${escapetool.xml($serverName)}"&gt;
+                      &lt;div class="filter-mme-server-name"&gt;${escapetool.xml($serverName)}&lt;span class="matches-count"&gt;&lt;/span&gt;&lt;span class="fa fa-info-circle mme-disclaimer-help" title="$escapetool.xml($services.localization.render('phenotips.myMatches.disclaimer.showDisclaimer'))"&gt;&lt;/span&gt;&lt;/div&gt;
```
marta-: Does this need to be a div? divs (block elements) inside labels are not valid markup.

matching-notification-ui/src/main/resources/MatchingNotification/MatchesTable.xml
```diff
@@ -127,20 +128,35 @@ $xwiki.jsx.use('PhenoTips.Widgets')##
...
+                      &lt;br&gt;
```
marta-: line breaks should be between label elements, not inside.

veronikaslc:
@marta- addressed codereview comments

veronikaslc:
approved

sashaandjic:
image

Everything works well. If the refresh was not successful an error message is presented to the user.

However, the drop down looks a bit like a big text box now. It may not be intuitive to the user that they can click on each item to refresh matches. 

I would add tooltip to both buttons with the following text:
- for REFRESH PATIENT MATCHES "Click here to refresh matches for PhenomeCentral and all MME members"
- for the arrow button "Click here to refresh matches for individual MME member"  

Would it be useful to add tooltip "Click here to refresh matches" when hovering over specific server in the dropdown? Or "Click here to refresh matches for" + <server name> ?

How about displaying server names in orange font?

sashaandjic:
@marta- ^

marta-:
Ok with tooltips (i.e. `title` attributes to be displayed on hover), but no "Click here" in the text, please. "Refresh matches for PhenomeCentral and all MME sites", " Choose MME site to refresh matches for" (is that proper English?).

Error mesages should be red. I'd leave server names black.

sashaandjic:
- tooltips are looking good
image

image

- problem: I can no longer see error messages in the dropdown list nor in the  Admin /Refresh Matches
image

I setup Australian Genomics Alliance with an invalid BASE URL in Administration: Remote Matching

allasm:
@sashaandjic this is expected, and is not a regression. We only store errors in the DB if a request was made (and possibly an error was returned). If the request never left the computer (because with a wrong URL there is nowhere to "leave") the error is not stored in the DB (and has never been). This can be improved but I think we decided is not worth it. Try a wrong MME key but correct URL for testing

sashaandjic:
tested
error msg shown properly
image

PR #296, opened 2019-08-15
PN-501: Improve the consistency within the matching table UI

veronikaslc:
Done.


matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.css
```diff
@@ -125,10 +125,6 @@
   margin-left: 20px;
 }
 
-#table-container {
-  margin-top: 1em;
-}
-
```
marta-: Why this change?

veronikaslc: removed duplicated property

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -937,7 +937,7 @@ var PhenoTips = (function (PhenoTips) {
         }
         // notes icon
         var icon = (match.notes) ? "fa fa-file" : "fa fa-file-o";
-        td += '<span class="buttonwrapper" title="' + this._NOTES_TITLE + '"><a class="button notes" href="#"><span class="' + icon + '"> </span></a></span>';
+        td += '<span class="notes ' + icon + '" title="' + this._NOTES_TITLE + '"> </span>';
```
marta-: This change and the one on line 926 must be accompanied by appropriate changes in the JS code which handles updating the icon when notes/comments are added, and hiding the popups. Currently, JS errors caused by outdated selectors (e.g. `elm.up('td').down('.notes span.fa').className = (textarea.value) ? "fa fa-file" : "fa fa-file-o";` where `'.notes span.fa'` no longer exists and should be replaced with `'span.fa.notes'`) are preventing the intended behavior of icon update and popup closing.

image

PR #297, opened 2019-09-13
PN-402: In matching notification table, when it is not clear which side should be contacted, "Contact" button should display a dialogue to choose the side to contact

allasm:


matching-notification-api/src/main/resources/ApplicationResources.properties
```diff
@@ -172,6 +172,10 @@ phenotips.myMatches.contact.dialog.to.label=To
 phenotips.myMatches.contact.dialog.from.label=From
 phenotips.myMatches.contact.dialog.subject.label=Subject
 phenotips.myMatches.contact.dialog.message.label=Message
+phenotips.myMatches.contact.dialog.selectPatientToContactTitle=Select which of the patients to contact
+phenotips.myMatches.contact.dialog.selectPatientToContactText=Please select which patient to contact
```
marta-: Nobody is contacting patients. Some possible suggestions:
- replace "patient" with "patient record owner"
- use "person of contact" , but then we'd have to use a different verb instead of "contact"
- rephrase the whole thing as a question: "who should be contacted"/"whom to contact"

sashaandjic:
My matches page does not load http://172.20.4.77:8080/MatchingNotification/MatchesTable

sashaandjic:
image

UI suggestions after talking to 2 native English speakers:
1. dialog title should be "Select contact"
2. modify text in the dialog to : "Please select a patient record owner to contact"
3. remove "Contact" and "about" from orange buttons labels and put the ID in brackets ( e.g. Mary Green (P0000002)) 
4. change "CANCEL CONTACT" to "CANCEL"

PR #299, opened 2019-09-20
PN-504: Create match similarity view widget for matches table; PN-479: If available add diagnosis for the solved cases on My Matches page

veronikaslc:


sashaandjic:
Another problem: once I added remote server My Matches page fails on open
image

sashaandjic:
On My Matches page the wrong matching details are shown
image

Patient form does not have this problem.

sashaandjic:
I made a new VM today http://172.20.4.131:8080 and did not see the problems previously reported. However it appears that incoming and outgoing matches appear identical in the matching table
image

veronikaslc:
@sashaandjic The issue with incoming and outgoing matches to be discussed separately.

sashaandjic:
It will look better if PatientIDs are aligned with details below
image

sashaandjic:
Marta and I looked at the ui and this is what needs to be done:
- align to the right patientIDs
- show matched and unmatched genes; use the same layout as in other sections (list genes for both patients)
- remove reference column
- not sure if we need MATCHED labels any more as we the top labels MATCH and CONTACT are sufficient
image

sashaandjic:
notes from me and @marta- 
- removed (cadidate), (VCF uploaded) and (FROM VCF) from all headings
image
(from VCF) can go together with variants info or not be shown at all at this stage.

- the border should be the same as in the notes, discussion and history ( thin black, rather than thick gray)

matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
```diff
@@ -345,6 +366,22 @@ public AccessLevel getAccess()
...
+        AccessLevel useAccess = entityAccess.getAccessLevel();
+        Visibility visibility = entityAccess.getVisibility();
+        if (useAccess.compareTo(MATCH) < 0 && visibility.compareTo(MATCHABLE_VISIBILITY) >= 0) {
+            // matches which are not matchable are filtered out elsewhere. But for MME requests which are
+            // done as a guest user (who has no access to all patients), need to make sure the match can
+            // be processed and returned, for which there should be at least "match" access level
+            useAccess = MATCH;
+        }
+
+        return new DefaultAccessType(this.access, VIEW, MATCH);
```
allasm: the local `useAccess` variable is never used? It is set but then the function does not use it anywhere and returns something which does nto depend on it

matching-notification-api/src/main/java/org/phenotips/matchingnotification/internal/DefaultMatchingNotificationManager.java
```diff
@@ -89,7 +89,7 @@
 
             PatientMatchNotificationResponse notificationResult = this.notifier.notify(email);
 
-            if (notificationResult.isSuccessul()) {
+            if (notificationResult != null && notificationResult.isSuccessul()) {
```
allasm: It looks like we did not test what happens when `notificationResult == null`. So this change is good. But what happens when we call `responses.add(notificationResult);` when notificationResult it is `null`, do we handle it correctly there?

matching-notification-api/src/main/java/org/phenotips/matchingnotification/match/internal/DefaultPatientInMatch.java
allasm: P.S. I wonder, do we still really need access stuff? We show matched genes either way, do we need it to see the variants?

matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchesTable.js
```diff
@@ -828,9 +841,7 @@ var PhenoTips = (function (PhenoTips) {
     _aggregatePhenotypes : function(match)
     {
         var matchPhenotypes = [];
-        var allPhenotypes = match.matched.phenotypes.predefined.concat(match.reference.phenotypes.predefined)
-                                                               .concat(match.reference.phenotypes.freeText)
```
allasm: Why did we remove that `.concat` - because phenotypes is now a different structure?

matching-notification-ui/src/main/resources/MatchingNotification/AdminMatchNotificationEmailTemplate.xml
```diff
@@ -249,7 +249,7 @@
       &lt;/ul&gt;
     #end
   #end
-  #if ($freetextFeatureSet.size() &gt; 0)
+  #if ($freetextFeatureSet &amp;&amp; $freetextFeatureSet.size() &gt; 0)
```
allasm: is `freetextFeatureSet` still relevant?

similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/RestrictedPatientSimilarityView.java
```diff
@@ -130,19 +127,6 @@ public DocumentReference getReporter()
...
-        return new RestrictedPatientPhenotypeSimilarityView(match, reference, access);
-    }
-
-    @Override
-    protected DisorderSimilarityView createDisorderSimilarityView(Disorder match, Disorder reference, AccessType access)
-    {
-        return new RestrictedDisorderSimilarityView(match, reference, this.access);
-    }
-
     @Override
```
allasm: if the only thing restricted is genes, do we need `RestrictedPatientSimilarityView` and not just `(restricted)GenotypeSimilarityView`?

allasm:
@sashaandjic @marta- "add unmatched genes" - even when patient is inaccessible to the viewing user? So we decided all genes (unlike variants) are OK to be seen? Just want an explicit statement on that

sashaandjic:
> @sashaandjic @marta- "add unmatched genes" - even when patient is inaccessible to the viewing user? So we decided all genes (unlike variants) are OK to be seen? Just want an explicit statement on that

Yes. We currently show unmatched genes on My Matches page in production.
My Matches page - typical user_LI

sashaandjic:
UI is fantastic. I found one problem: P0000002 has two genes, P0000007 none. The simmilarity view should show P0000002 genes under unmatched category.
image

similarity-genotype-view/src/main/java/org/phenotips/data/similarity/genotype/ExomiserExome.java
```diff
@@ -96,10 +96,6 @@
             }
 
             geneMutations.add(variant);
-            // Represent homozygous variants as two separate mutations
```
allasm: any reason to change this logic? do you know why we had it like this in the first place?

marta-: Clearly labeling the variant as "homozygous" in the UI is easier for the user to process than duplicating the variant and letting the user notice there are two identical entries.

The duplication of homozygous variants was introduced by @buske [6 years ago](https://github.com/phenotips/patient-network/blame/f4cef44d8389e62365651fb2823748c94c8a364e/similarity-data-impl/src/main/java/org/phenotips/data/similarity/internal/ExomizerGenotype.java#L81).

buske: Oh goodness, yes, it was the easiest approach at the time... sorry.

sashaandjic:
@veronikaslc local matches are again in a weird state. In the sample below, the similarity view is showing  P0000017 instead of P0000018.
image

sashaandjic:
Another problem: When in patient form, all filters need to be applied against Matched side.
For example I want to see all P0000001 marches that have specific gene. Right now it filters both sides, so if P0000001 happens to have this gene, all its matches will be shown. 

Also, both MATCHED labels are still present in the headers' second row. They should be removed.

sashaandjic:
Clinical diagnosis of the matching patient isn't shown in the similarity view.
Everything else works as expected.

sashaandjic:
Free text diagnoses and phenotypes are not taken into account. Please see patient P0000002.

veronikaslc:
@sashaandjic @marta- Free text features were not included in the patient similarity views by design. Do we still want to show them?

sashaandjic:
Currnetly free text in genes is supported. We should either support free text for all or for none.
My vote goes to all. 
image

sashaandjic:
free text phenotypes are still missing
image

sashaandjic:
The only complain that I have is that unlike free text genes , we do not match free text phenotypes
image

sashaandjic:
for this remote match phenotype Intellectual disability isn't shown as matching phenotype (both patients have it)
image

sashaandjic:
filtering does not work for free text phenotypes (it works for genes)
image

PR #300, opened 2019-09-24
PN-505: PN-456 migration fails in production environment

allasm:
Fixed


allasm:
Supposedly with newer hibernate escaping `:` with `\\` should work, but it does not for us (not sure if Hibernate is old or something else). There is a more clean way which is to use `doWork()` to run raw SQL queries, but it requires many more changes and not sure it would be better for this one-time hotfix

PR #303, opened 2019-10-31
[misc] Parse remote patient ID in the match details dialog

veronikaslc:


matching-notification-resources/src/main/resources/resources/uicomponents/matchingNotification/matchDetailsView.js
```diff
@@ -51,8 +51,19 @@ var PhenoTips = (function(PhenoTips) {
...
-        row.insert(new Element('th', {'class' : 'patient-name result'}).update(r.matched.patientId + matchedExternalId));
+        row.insert(new Element('th', {'class' : 'patient-name'}).update(this._getRemotePatientId(r.reference.patientId, r.reference.serverId) + refExternalId));
+        row.insert(new Element('th', {'class' : 'patient-name result'}).update(this._getRemotePatientId(r.matched.patientId, r.matched.serverId) + matchedExternalId));
+    },
+
+    _getRemotePatientId : function(patientId, serverId) {
+        // Parse remote patient ID because MyGene2 uses URLs in the ID field to ensure that the public MyGene2 profiles are actually delivered to the end user
+        if (serverId == "mygene2" && patientId && patientId.startsWith("http")) {
+            var matches = patientId.match(/[\d]+/g);
+            var id = matches ? matches[matches.length - 1] : "$escapetool.javascript($services.localization.render('phenotips.similarCases.profile'))";
```
sdumitriu: Why fall back to `phenotips.similarCases.profile`? In the unlikely case that the URL doesn't match the pattern, shouldn't we instead use the whole URL?

